PR #4, opened 2022-08-18
feat(harvester): implement JFR harvester and push-to-archives

### LINKED ISSUE #22: [Task] Start JFR dynamically by configuration
The Agent should be able to start Flight Recordings using a `.jfc` template name and other settings (`maxSize`/`maxAge`) by picking up configuration properties.

### LINKED ISSUE #24: [Task] Implement scheduled periodic recording uploads
The Agent should expose configuration properties for periodic pushes of captured JFR data to the Cryostat backend. After the Agent has successfully registered with Cryostat and started a Flight Recording locally, it should periodically push the current JFR buffer contents to Cryostat as a full `.jfr` file. In the future there can also be work done to stream the data, perhaps on a per-event basis or at least on a per-chunk basis, but that is out of scope for this initial implementation.

andrewazores:
Depends on https://github.com/cryostatio/cryostat-core/pull/169
Depends on #18

Related to https://github.com/cryostatio/cryostat/issues/520
Related to https://github.com/cryostatio/cryostat/issues/1013

Fixes #22
Related to #23 - there is still experimentation and work to be done to expose controls for the emergency dump data age and filesize
Fixes #24
See also #25

How to test: usual `-agent` test setup in brief
1. Check out PR, `mvn clean install` to get agent JAR into local `~/.m2`
2. Check out latest `quarkus-test`, `./mvnw -U clean package && podman build -t quay.io/andrewazores/quarkus-test:latest -f src/main/docker/Dockerfile.jvm . ; podman image prune -f` to build a `quarkus-test` with the PR agent included
3. `cd cryostat`, edit one of the `quarkus-test-agent` sample apps to use `:latest` tag from image above
4. `sh smoketest.sh`

After that, the standard `smoketest.sh` should be running with Cryostat working as usual, three `vertx-fib-demo` as usual, and two `quarkus-test-agent` - one with the PR version agent and one with the `main` version.

`podman logs -f quarkus-test-agent-1` (if `-1` is the one with the PR agent) to check what the new agent says it's doing. With no additional configuration done, the new and old agents should behave as usual - only as discovery plugins, with no harvester behaviour. They should be visible in the Cryostat API in all the usual ways.

Edit `smoketest.sh`:

```diff
diff --git a/smoketest.sh b/smoketest.sh
index 48c0f88c..28937c59 100755
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -126,6 +126,7 @@ runDemoApps() {
         --env QUARKUS_HTTP_PORT=10010 \
         --env ORG_ACME_CRYOSTATSERVICE_ENABLED="false" \
         --env CRYOSTAT_AGENT_APP_NAME="quarkus-test-agent" \
+        --env CRYOSTAT_AGENT_HARVESTER_PERIOD_MS=30000 \
         --env CRYOSTAT_AGENT_WEBSERVER_HOST="localhost" \
         --env CRYOSTAT_AGENT_WEBSERVER_PORT="9977" \
         --env CRYOSTAT_AGENT_CALLBACK="http://localhost:9977/" \
```

to enable the harvester behaviour on one of the `quarkus-test` sample apps, and re-run `smoketest.sh`. This configures the agent to start a recording (defaults to using the `default.jfc` shipped with the host JVM) and to push the collected data to the Cryostat backend every 30 seconds. These should be visible in the Cryostat UI as they should produce upload notifications, and should be listed in the `Archives > Uploads` view.


andrewazores:
This needs some cleanup around the build setup and JDK 11 vs 17, etc. Manually building things locally with JDK 17 across all projects should work as expected however.

andrewazores:
Bah. I have a hunch that this CI setup to both pull authenticated from -core and push authenticated to -agent is what's causing the pain above, since these are two separate project repositories with their own unique tokens that do not grant each other permissions to pull from each others' package repos, only to push builds to their own package repos.

andrewazores:
Interesting, the CI shows as passing now. I think it's getting confused because the other PR #19 has an identical set of commits but on an upstream branch. Before I opened that other PR this one showed as failing because of a 401 when trying to authenticate to the GH packages repo to download -core.

andrewazores:
#20 also worked, I think what's happening is that the -core dependency got cached by #19's successful build, and so these subsequent PR rebuilds are also seeing that cached artifact and so mvn doesn't try to connect to the GitHub Packages repo, so the lack of valid auth token doesn't break anything.

andrewazores:
Replaced by #26 - same commits, but on an upstream branch for to pass.

PR #7, opened 2022-08-25
feat(discovery): generate UUID in self node

andrewazores:
Generate a UUID for the self-referencing discovery node

Related to https://github.com/cryostatio/cryostat/pull/1056
Depends on #12


andrewazores:
Obsolete

PR #31, opened 2023-01-13
ci(dependabot): add config

andrewazores:
Fixes #14

PR #32, opened 2023-01-13
ci(dependabot): add config

andrewazores:
Replaces #31

### DIFF
--- a/src/main/java/io/cryostat/agent/MainModule.java
+++ b/src/main/java/io/cryostat/agent/MainModule.java
@@ -37,7 +37,6 @@
  */
 package io.cryostat.agent;
 
-import java.io.IOException;
 import java.net.URI;
 import java.net.http.HttpClient;
 import java.security.KeyManagementException;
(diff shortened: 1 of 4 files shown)


src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -231,15 +226,13 @@ public void print(String msg) {
...
-                | MBeanException
-                | InstanceNotFoundException
-                | AttributeNotFoundException e) {
+        try {
+            try (JFRConnection connection = tk.connect(tk.createServiceURL("localhost", 0))) {
+                String id = connection.getJvmId();
+                log.info("Computed self JVM ID: {}", id);
+                return id;
+            }
+        } catch (Exception e) {
```
ebaron, line 235: Is this nested `try` intentional?

andrewazores: Unfortunately yes, the implicit `connection.close()` on the try-with-resources can actually throw on its own too, so the outer try is to catch from that.

ebaron: Ah, I see. Thanks for the info!

ebaron (approved):

Issue #63, opened 2023-02-21
[Task] Implement Basic authentication on requests


andrewazores:
Depends on https://github.com/cryostatio/cryostat/issues/1376

PR #64, opened 2023-02-21
feat(plugin): require HTTP Basic authentication

andrewazores:
Fixes https://github.com/cryostatio/cryostat-agent/issues/63
Depends on https://github.com/cryostatio/cryostat/pull/1377 (so the backend knows how to pass the challenge)

~~In this PR the Agent generates Basic auth credentials to secure its built-in HTTP webserver. Incoming HTTP requests to the webserver must pass an HTTP Basic Authentication challenge. The credentials use a fixed username and a randomly generated password string. The password string is held in memory only long enough for the Agent to complete registration and publication of itself to the Cryostat backend. After that point, the password is blanked out from memory and only a hash of the password retained to compare against incoming HTTP requests. This PR also causes the Agent to prefer to register itself with the HTTP Agent connection URL rather than a JMX Service URL, but there is a config property to prefer using a JMX Service URL (contingent on JMX being detected enabled on the host JVM).~~

Updated: https://github.com/cryostatio/cryostat/pull/1377#issuecomment-1444329903
The username and password generation and clearing behaviour is still the same, however, rather than passing these details in the callback URI userinfo, they are stored in the Cryostat server's encrypted credentials database. There is a userinfo reference format that specifically tells Cryostat to look up the credentials in that database rather than reading them literally from the URI.

This is a precursor to expanding the Agent HTTP interface to implement a subset of Cryostat API actions, ex. retrieving the list of active recordings, dynamically starting recordings, etc. These actions should obviously require clients to pass an authentication challenge.

### DIFF
--- a/src/main/java/io/cryostat/agent/Registration.java
+++ b/src/main/java/io/cryostat/agent/Registration.java
@@ -69,6 +69,7 @@ class Registration {
     private final String realm;
     private final String hostname;
     private final URI callback;
+    private final boolean preferJmx;
     private final int jmxPort;
     private final int registrationRetryMs;
 
@@ -83,6 +84,7 @@ class Registration {
             String realm,
             String hostname,
             URI callback,
+            boolean preferJmx,
             int jmxPort,
             int registrationRetryMs) {
         this.executor = executor;
@@ -92,6 +94,7 @@ class Registration {
         this.realm = realm;
         this.hostname = hostname;
         this.callback = callback;
+        this.preferJmx = preferJmx;
         this.jmxPort = jmxPort;
         this.registrationRetryMs = registrationRetryMs;
     }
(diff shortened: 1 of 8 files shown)


src/main/java/io/cryostat/agent/Registration.java
```diff
@@ -192,19 +229,25 @@ private DiscoveryNode defineSelf() throws UnknownHostException {
...
                                     "service:jmx:rmi:///jndi/rmi://%s:%d/jmxrmi",
                                     hostname, jmxPort));
-        } else {
-            uri = callback;
         }
         DiscoveryNode.Target target =
                 new DiscoveryNode.Target(
-                        realm, uri, appName, jvmId, pid, hostname, jmxPort, javaMain, startTime);
+                        realm,
+                        callback,
```
ebaron, line 243: Just want to double-check that this should be `callback` and not `uri`

ebaron (commented):
Hey Andrew, looks good. Just one question.

ebaron (approved):

src/main/java/io/cryostat/agent/Registration.java
andrewazores: Ah, I think you're right. I'm rebuilding to test it now but this makes sense.

ebaron (approved):

andrewazores:
Ah, thanks to CI there's a Java 11 compatibility issue uncovered.

andrewazores:
@ebaron ping, I had to make a small touchup in the last commit. My system is normally building with JDK 17 so nothing complained when I used some more recent API from the JDK webserver package that isn't actually there in 11 which is our real build target. Thankfully it was quite straightforward to match back up to the older API, in fact this change is even better because it lets me track the elapsed time for the request.

ebaron (approved):

Issue #14, opened 2022-11-08
[Task] Set up Dependabot


andrewazores:
See https://github.com/cryostatio/cryostat-core/issues/168 . Something similar should be done here.

andrewazores:
Fixed by #32

PR #65, opened 2023-02-22
ci(deps): configure Dependabot to read from GH packages

ebaron:
To handle cryostat-core updates, as done in: https://github.com/cryostatio/cryostat-reports/pull/76.

Related to: #14

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,10 @@
 version: 2
+registries:
+  maven-github:
+    type: maven-repository
+    url: https://maven.pkg.github.com/cryostatio/cryostat-core
+    username: dummy
+    password: ${{secrets.DEPENDABOT_READ_GHCR}}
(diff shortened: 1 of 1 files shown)


andrewazores:
I think this is currently failing because the CI runner probably doesn't have access to the required secret. I believe it would need to run as a `pull_request_target` event to read the secret token.

ebaron:
> I think this is currently failing because the CI runner probably doesn't have access to the required secret. I believe it would need to run as a `pull_request_target` event to read the secret token.

I think so too. I ran into the same issue for -reports.

tthvo (approved):
Looks good! I think dependabot will be opening PR as upstream branch so this should pass as expected?

andrewazores:
Yup, that's true. There is a new -core version that I just published a few minutes ago so we should see an update soon, hopefully.

PR #86, opened 2023-03-23
fix(registration): registration checks and re-registration flow

### LINKED ISSUE #85: [Bug] Agent registration flow after Cryostat restarts
The Agent's startup registration flow is as follows:

1. Start up, check all required environment variables. Fail if any not defined.
2. Generate authorization credentials for own webserver
3. Create a matchExpression tailored to itself (`target.jvmId == '$myHashId'`), publish a stored credential to Cryostat using this expression and the generated credentials. Retry periodically until this succeeds.
4. Register as a discovery plugin. Retry periodically until this succeeds.
5. Publish a description of itself as the single node under its plugin registration.

This is mostly resilient on the Cryostat side against various conditions where communications between the Agent and Cryostat break down, or Cryostat goes down temporarily, because there is the callback URI ping mechanism, where Cryostat will attempt to call the plugin back to 1. ensure that it is still there, 2. ensure the plugin (agent) refreshes its registration token before expiry, 3. Cryostat's stored credentials for the plugin are still valid.

However, this can be problematic on the agent side in the case that Cryostat goes down temporarily and the agent stays up. In this case, Cryostat will call back the plugin with the old callback and old stored credentials, which will fail, prompting Cryostat to deregister the plugin. Cryostat will then forget that this plugin ever existed and not try to ping its callback anymore. Meanwhile, the Agent considers itself registered, published, and up-to-date, and will happily sit idle forever, never renewing its registration with Cryostat. If the user stops the agent (its attached application) and restarts it then the flow will properly resume from the top and everything should be properly registered and published again, however, a properly resilient system would reach this state again when either Cryostat or the agent go down temporarily and reappear later.

Moreover, if the Agent does notice that it is no longer registered with Cryostat and attempts to re-register itself, it will fail at Step 3, because it will regenerate an identical matchExpression, which Cryostat will reject because it expects matchExpressions to be unique. Cryostat should communicate this specific failure mode (ex. via response status code) in a way that the agent recognizes so that it can continue to step 4. Other failure modes should still block progression and probably lead to retrying step 3.

andrewazores:
- fix(registration): check if credentials already exist before submitting
- fix(registration): check and retry registration on slow cadence
- extract registration (re)check config property

Fixes #85
Depends on https://github.com/cryostatio/cryostat/pull/1425

### DIFF
--- a/src/main/java/io/cryostat/agent/Agent.java
+++ b/src/main/java/io/cryostat/agent/Agent.java
@@ -96,12 +96,10 @@ public class Agent {
                     evt -> {
                         switch (evt.state) {
                             case REGISTERED:
-                                break;
                             case UNREGISTERED:
-                                break;
                             case REFRESHED:
-                                break;
                             case PUBLISHED:
+                                log.info("Registration state: {}", evt.state);
                                 break;
                             default:
                                 log.error("Unknown registration state: {}", evt.state);
--- a/src/main/java/io/cryostat/agent/Agent.java
+++ b/src/main/java/io/cryostat/agent/Agent.java
@@ -106,7 +106,8 @@ public class Agent {
                                 break;
                         }
                     });
-            webServer.start().thenRun(registration::start).get();
+            webServer.start();
+            registration.start();
             log.info("Startup complete");
         } catch (Exception e) {
             log.error(Agent.class.getSimpleName() + " startup failure", e);
--- a/src/main/java/io/cryostat/agent/Agent.java
+++ b/src/main/java/io/cryostat/agent/Agent.java
@@ -96,10 +96,8 @@ public class Agent {
                     evt -> {
                         switch (evt.state) {
                             case REGISTERED:
-                                harvester.start();
                                 break;
                             case UNREGISTERED:
-                                harvester.stop();
                                 break;
                             case REFRESHED:
                                 break;
--- a/src/main/java/io/cryostat/agent/Agent.java
+++ b/src/main/java/io/cryostat/agent/Agent.java
@@ -110,8 +110,7 @@ public class Agent {
                                 break;
                         }
                     });
-            webServer.start();
-            registration.start();
+            webServer.start().thenRun(registration::start).get();
             log.info("Startup complete");
         } catch (Exception e) {
             log.error(Agent.class.getSimpleName() + " startup failure", e);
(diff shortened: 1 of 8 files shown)


maxcao13 (commented):
What is the intended result if the user deletes the jvmId==hash credential within Cryostat? After a non-200 GET response from the check, the agent will repeatedly try to POST another registration request many many times in quick succession until something happens in the end that I can't understand.

Here is the attached logs, but it seems both agent worker threads POST to the discovery endpoint during the same time and the retries stop. Then we must sigkill that agent.

[t.log](https://github.com/cryostatio/cryostat-agent/files/11065556/t.log)

andrewazores:
Hmm, I'll try to figure out what's happening there. If the stored credentials get deleted then that shouldn't actually do anything to the agent right away - its registration information and token are still valid, so it should still be able to check its registration status periodically. If it needed to it should still be able to publish updates about its discovery subtree (just itself) as well, but the timing there is very unlikely.

However, once the credentials are deleted, then when Cryostat tries to ping the agent instance after some time, the ping should fail because Cryostat is unable to look up the stored credentials that the agent expects, resulting in the agent rejecting Cryostat's ping. This should lead to Cryostat deregistering the agent. The next time the agent tries to check its registration state it will see this and should re-enter its initial registration flow, which should start with attempting to re-submit new stored credentials.

andrewazores:
I think I've solved it with the last commit. @maxcao13 give it another test run please

maxcao13:
I tried the pr and the issue with the looping registration is fixed now, but I notice something else weird now.

I tried lowering the ping period of the discovery storage pinging the plugins `CRYOSTAT_DISCOVERY_PING_PERIOD="10000" for 10 seconds, and it seems like the harvester for `quarkus-test-agent-1` runs into some issues with not being able to upload the exit dump file.

andrewazores:
Ah yes, I think that makes sense with the way the agent's internal state machinery works. The discovery ping request tells the agent to re-register itself, which really means de-register and start the registration flow over. But, the harvester also intentionally stops itself when the agent has become deregistered, and starts on successful registration. So if the discovery ping period is shorter than the harvester period you'll end up with the agent probably never even attempting to push harvested files, or if it does then perhaps it gets interrupted part way through sometimes.

There's probably some approach I could take like adding an internal state for "re-registering" to allow this to smoothly transition over, but I'm not sure how worthwhile it is.

This does point out an interesting side-effect of the discovery ping though, because it means that the harvester period will result in files being pushed with that periodicity on one interval, but the interval will reset at the next re-registration time. If the discovery ping period is relatively long and the harvester period is relatively short then this won't be very noticeable, but as these two periods become closer in value then the skew could become noticeable, up to the point you've identified where the skew actually can prevent any harvesting from happening at all.

I think that's a separate issue to work on, maybe for next development cycle.

maxcao13 (dismissed):
Looks good to me!

andrewazores:
Rebased

maxcao13 (approved):

PR #100, opened 2023-04-15
fix(states): add REFRESHING state to avoid deregistration skew and cancellations

andrewazores:


tthvo (approved):
Looks good to me! Lately, I have been seeing the agent disappearing and re-appearing frequently. Looks like its fixed here too^^

andrewazores:
Yea, that's the other side of the same problem I was looking to fix here. The description of the agent fully deregistering and re-registering when it gets the ping signal not only causes its harvest timer to reset, but it also means the Cryostat server sees the Agent's discovery subtree (singleton of only itself) being removed, then re-created empty, and finally an identical node re-appearing as the singleton leaf.

PR #134, opened 2023-05-30
fix(credentials): store credentials with non-conflicting matchExpression

### LINKED ISSUE #133: [Story] Agent credentials storage conflicts with authenticated JMX
The Agent generates its HTTP webserver credentials and then stores them in the Cryostat server's encrypted keyring along with the other stored credentials. Cryostat consults these stored credentials and their matchExpressions to determine what credentials to pass when initiating a JMX connection. If the Agent is installed into a target application that is configured with JMX authentication, then the Agent will store credentials that also match the target's JMX definition, since the matchExpression used by the agent is `target.jvmId == 'abcd1234'`. This conflict results in Cryostat failing to establish a JMX connection to the target since the Agent's webserver credentials are incorrectly passed.

andrewazores:
Fixes #133

This simply changes the matchExpression that the Agent uses when it stores credentials in the Cryostat server's encrypted keyring.

Previously the matchExpression used would be set to match the target's JVM ID. As outlined in the original issue, this would cause problems for target applications with JMX authentication enabled, since the credentials stored by the Agent would match the same target, but are the wrong credentials for that connection.

It does not actually matter what the matchExpression of these stored credentials is because there is a custom subprotocol in use by the Agent and the Cryostat server, so the Cryostat server will always use the correct stored credentials given the correct ID for the Agent it is trying to connect to over HTTP. The matchExpression is simply used in this case to provide a meaningful link between the stored credential and the agent instance so that the user can tell which credentials are used for which connections.

The new matchExpression matches on the specific callback URL for the Agent instance. This should uniquely identify the Agent instance for the user, and also prevents the credential from being incorrectly picked up and used for any JMX connections that Cryostat initiates.

`quay.io/andrewazores/vertx-fib-demo:0.12.3` has been updated to use this modified agent, so its behaviour can be easily observed by updating the `smoketest.sh` script to update that image version. No other changes should be required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 <modelVersion>4.0.0</modelVersion>
 <groupId>io.cryostat</groupId>
 <artifactId>cryostat-agent</artifactId>
-<version>0.3.0-SNAPSHOT</version>
+<version>0.2.3</version>
 <packaging>jar</packaging>
 <name>cryostat-agent</name>
 <url>http://maven.apache.org</url>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 <modelVersion>4.0.0</modelVersion>
 <groupId>io.cryostat</groupId>
 <artifactId>cryostat-agent</artifactId>
-<version>0.2.3</version>
+<version>0.3.0-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


tthvo (approved):
Looks good!

PR #149, opened 2023-06-13
ci(secrets): allow ci to use gh package secrets

### LINKED ISSUE #126: [Task] CI fails to download `cryostat-core` dependency
For example:

https://github.com/cryostatio/cryostat-agent/pull/125
(https://github.com/cryostatio/cryostat-agent/actions/runs/5070291951/jobs/9105123984?pr=125)
image

The project's `pom.xml` is configured to search the GitHub Maven Packages repository, but authentication is failing.

https://github.com/cryostatio/cryostat-agent/blob/8d2eb414029c436efde2c9b2d0a308d0dad59d66/pom.xml#L13

The CI workflow does have a step that is supposed to set up the Maven `settings.xml` that provides the credentials however:

https://github.com/cryostatio/cryostat-agent/blob/8d2eb414029c436efde2c9b2d0a308d0dad59d66/.github/workflows/ci.yaml#L32

And the same or very similar configuration is used in the main Cryostat repository, which works and is able to download the `cryostat-core` dependency:

https://github.com/cryostatio/cryostat/blob/2b0cd5adacf8f928f8666f36663b0a5b31e0e76b/.github/workflows/ci-jobs.yml#L53

maxcao13:
Fixes #126

Not sure if it works because my fork doesn't own a gh repo token that is authenticated for use for pulling a `-core` dependency - does the fix make sense?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,7 +23,8 @@ jobs:
       run: exit 1
     - uses: actions/checkout@v2
       with:
-        repository: cryostatio/cryostat-core
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.event.pull_request.head.ref }}
         submodules: true
     - uses: actions/setup-java@v2
       with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,13 +15,27 @@ on:
       - cryostat-v[0-9]+.[0-9]+
 
 jobs:
-
+  get-pom-properties:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Fail if safe-to-test label NOT applied
+      if: ${{ !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
+      run: exit 1
+    - id: query-pom
+      name: Get properties from POM
+      run: |
+        CORE_VERSION="$(mvn help:evaluate -Dexpression=io.cryostat.core.version -q -DforceStdout)"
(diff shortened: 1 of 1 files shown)


ebaron (commented):
This makes sense, but we'll want to guard against unauthorized access to secrets using the `safe-to-test` label used in other repos:
https://github.com/cryostatio/cryostat/blob/7ec485552750296463d19ad9075347badc60c312/.github/workflows/pr-ci.yml#L29

We'll also need to revise the repo and ref that gets checked out for PRs:
https://github.com/cryostatio/cryostat-operator/blob/1a02ba82282cec37931b9993970dc416c0e1d08a/.github/workflows/ci.yaml#L45-L46

andrewazores (approved):

andrewazores:
Whoops:

https://github.com/cryostatio/cryostat-agent/actions/runs/5279418179/jobs/9550060578

The `safe-to-test` label needs to be checked only for PR CI runs, not on push (merge) to `main`.

This repo doesn't currently have a split definition for push-ci vs pr-ci, so I think it'll need that done too.

ebaron:
> Whoops:
> 
> https://github.com/cryostatio/cryostat-agent/actions/runs/5279418179/jobs/9550060578
> 
> The `safe-to-test` label needs to be checked only for PR CI runs, not on push (merge) to `main`.
> 
> This repo doesn't currently have a split definition for push-ci vs pr-ci, so I think it'll need that done too.

This is how @tthvo worked around it in the operator CI: https://github.com/cryostatio/cryostat-operator/blob/1a02ba82282cec37931b9993970dc416c0e1d08a/.github/workflows/ci.yaml#L41

PR #156, opened 2023-06-20
fix(credentials): store credentials with unique matchExpression

### LINKED ISSUE #154: [Bug] Server stored credentials should include JVM ID for uniqueness
In #133 , the previous stored credentials match expression used only the target's JVM ID, which caused a conflict with credentials actually stored for JMX connection use. The new way uses the target's callback URL as the `target.connectUrl`, which will be an agent HTTP API URL and not conflict with JMX credentials. However, in some scenarios like unclean shutdowns of the Agent and/or Cryostat, old stale Credentials entries may be left behind in the encrypted keyring, which may have the same `connectUrl` as the (potentially restarted, or at least re-registering) Agent instance. This will cause another conflict and prevent Agent registration. To help mitigate this, Agents can store their credentials with an expression matching on both the callback/connectUrl as well as the JVM ID.

andrewazores:
Fixes #154


tthvo (approved):
Looks good to me!

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -58,6 +59,7 @@ public abstract class ConfigModule {
 
     private static final Logger log = LoggerFactory.getLogger(ConfigModule.class);
 
+    public static final String CRYOSTAT_AGENT_INSTANCE_ID = "cryostat.agent.instance-id";
```
tthvo, line 62: Document this new config in README?

src/main/java/io/cryostat/agent/CryostatClient.java
```diff
@@ -354,7 +360,10 @@ private CountingInputStream getRecordingInputStream(Path filePath) throws IOExce
     }
 
     private String selfMatchExpression(URI callback) {
-        return String.format("target.connectUrl == \"%s\"", callback);
+        return String.format(
+                "target.connectUrl == \"%s\" && target.jvmId == \"%s\" &&"
+                        + " target.annotations.platform[\"INSTANCE_ID\"] == \"%s\"",
```
tthvo, line 365: Any reason to use platform but not `cryostat` annotations?

andrewazores: https://github.com/cryostatio/cryostat/blob/bee2ffbc21c21fb231608e6c02e401fa0e951313/src/main/java/io/cryostat/platform/ServiceRef.java#L164

https://github.com/cryostatio/cryostat/blob/bee2ffbc21c21fb231608e6c02e401fa0e951313/src/main/java/io/cryostat/platform/ServiceRef.java#L210

The `cryostat` annotations are meant to have only this restricted key set, none of which really match what this is doing. This agent instance ID is somewhat similar conceptually to the hash that k8s uses to make ex. Pods unique, and those would also go in the `platform` annotations, so it makes sense to put it here from both angles IMO.

tthvo (approved):
Looks good to me!

maxcao13:
Is it me, or do I still see the old format: 

image

andrewazores:
In the `smoketest.sh` there are `quarkus-test` and `vertx-fib-demo` that both have Cryostat Agent instances on them, so both of those container images would need to be updated to include these changes. There are three of the `vertx-fib-demos` so that probably matches the three old-style credentials you're seeing.

tthvo:
Just wondering it is also because the database is not pruned off those entries? I am seeing all formats

image

maxcao13:
I figured out the problem. I was running `sh smoketest.sh` right after a `mvn clean package` in the base cryostat and I realized that the package lifecycle no longer builds an image with the tag `2.4.0-snapshot` anymore, and that's what `run.sh` defaults to based on the `pom.xml` if `CRYOSTAT_IMAGE` isn't set.

PR #157, opened 2023-06-20
fix(registration): correctly annotate with JMX port number

### LINKED ISSUE #153: [Bug] Discovery self-node has wrong annotation port number when publishing JMX URL
In `Registration.java`:

```java
        URI uri = callback;
        if (preferJmx && jmxPort > 0) {
            uri =
                    URI.create(
                            String.format(
                                    "service:jmx:rmi:///jndi/rmi://%s:%d/jmxrmi",
                                    hostname, jmxPort));
        }
        DiscoveryNode.Target target =
                new DiscoveryNode.Target(
                        realm,
                        uri,
                        appName,
                        jvmId,
                        pid,
                        hostname,
                        uri.getPort(),
                        javaMain,
                        startTime);
```

The `uri.getPort()` here returns the correct expected port number when the registration `connectUrl` is the agent's own HTTP connection. When the registration uses JMX, however, (ie when `registration.prefer-jmx` config property is set), then this returns `-1` instead of the actual JMX port number.

andrewazores:
Fixes #153

### DIFF
--- a/src/main/java/io/cryostat/agent/Registration.java
+++ b/src/main/java/io/cryostat/agent/Registration.java
@@ -291,12 +291,14 @@ class Registration {
                         .orElse(Instant.EPOCH)
                         .getEpochSecond();
         URI uri = callback;
+        int port = uri.getPort();
         if (preferJmx && jmxPort > 0) {
             uri =
                     URI.create(
                             String.format(
                                     "service:jmx:rmi:///jndi/rmi://%s:%d/jmxrmi",
                                     hostname, jmxPort));
+            port = jmxPort;
         }
         DiscoveryNode.Target target =
                 new DiscoveryNode.Target(
@@ -306,7 +308,7 @@ class Registration {
                         jvmId,
                         pid,
                         hostname,
-                        uri.getPort(),
+                        port,
                         javaMain,
                         startTime);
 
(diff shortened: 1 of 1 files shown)


tthvo (approved):
Looks good! Fixed the issue (`PORT` annotation?). 

Wondering why previously it does not behave correctly as it seemed to work?

andrewazores:
```
$ jshell
|  Welcome to JShell -- Version 17.0.7
|  For an introduction type: /help intro

jshell> URI.create("service:jmx:rmi:///jndi/rmi://localhost:1234/jmxrmi")
$1 ==> service:jmx:rmi:///jndi/rmi://localhost:1234/jmxrmi

jshell> URI.create("service:jmx:rmi:///jndi/rmi://localhost:1234/jmxrmi").getPort()
$2 ==> -1

jshell> URI.create("service:jmx:rmi:///jndi/rmi://localhost:1234/jmxrmi").getSchemeSpecificPart()
$3 ==> "jmx:rmi:///jndi/rmi://localhost:1234/jmxrmi"

```

The JMX service URL format's port number is not actually [the URI's](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URI.html) port number, but embedded within the "scheme-specific part" of the URI. So, the regular `.getPort()` method simply returns `-1` instead of what would be expected as the JMX port number. The rest of everything in this call chain works fine and puts the value into the right place, I just overlooked that this value will be wrong in the "prefer JMX" case.

tthvo:
Ahh okayy that makes sense! Thanks for explaining!

PR #165, opened 2023-07-05
feat(api): enable dynamic JFR start

andrewazores:
Fixes #163
Fixes #164
Depends on https://github.com/cryostatio/cryostat-core/pull/245

Not yet implemented: restarting already existing recordings, manual stop of running recordings, opening stream (download) recordings. This only enables the ability to remotely start a recording over HTTP.

Issue #124, opened 2023-05-23
[Epic] Two-way communications protocol


andrewazores:
The agent currently exposes only a readonly HTTP API that the Cryostat server can use to query basic information such as a list of active recording descriptors, the JFR event types and templates available, and MBean metrics data.

The agent should also implement mutation requests for actions such as dynamically starting Flight Recordings, including with custom event templates supplied by the server with the request.

The readonly API should remain implemented and available. We should explore whether the mutable (write/update) API endpoints are always available, or if they are gated behind an additional opt-in property that requires the user to explicitly enable them.

The goal after this project is completed is that [the Cryostat backend's client](https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/AgentConnection.java) should be able to make all of its requests to an HTTP-registered Agent instance equally as it would for a JMX target.

- [x] #138
- [x] #162
- [x] #178

andrewazores:
Related: https://github.com/cryostatio/cryostat-web/issues/944

andrewazores:
https://github.com/cryostatio/cryostat-agent/issues/138 should still be done at some point.

PR #176, opened 2023-07-31
feat(api): enable dynamic JFR stop, delete

andrewazores:
See #165
Related to #124

Adds remote endpoint ability to HTTP request Agent to stop or delete JFR by ID.

### DIFF
--- a/src/main/java/io/cryostat/agent/WebServer.java
+++ b/src/main/java/io/cryostat/agent/WebServer.java
@@ -163,7 +163,8 @@ class WebServer {
                 handler.handle(x);
             } catch (Exception e) {
                 log.error("Unhandled exception", e);
-                x.sendResponseHeaders(HttpStatus.SC_INTERNAL_SERVER_ERROR, 0);
+                x.sendResponseHeaders(
+                        HttpStatus.SC_INTERNAL_SERVER_ERROR, RemoteContext.BODY_LENGTH_NONE);
                 x.close();
             }
         };
@@ -184,15 +185,18 @@ class WebServer {
                     case "POST":
                         synchronized (WebServer.this.credentials) {
                             executor.execute(registration.get()::tryRegister);
-                            exchange.sendResponseHeaders(HttpStatus.SC_NO_CONTENT, -1);
+                            exchange.sendResponseHeaders(
+                                    HttpStatus.SC_NO_CONTENT, RemoteContext.BODY_LENGTH_NONE);
                         }
                         break;
                     case "GET":
-                        exchange.sendResponseHeaders(HttpStatus.SC_NO_CONTENT, -1);
+                        exchange.sendResponseHeaders(
+                                HttpStatus.SC_NO_CONTENT, RemoteContext.BODY_LENGTH_NONE);
                         break;
                     default:
                         log.warn("Unknown request method {}", mtd);
-                        exchange.sendResponseHeaders(HttpStatus.SC_METHOD_NOT_ALLOWED, -1);
+                        exchange.sendResponseHeaders(
+                                HttpStatus.SC_METHOD_NOT_ALLOWED, RemoteContext.BODY_LENGTH_NONE);
                         break;
                 }
             } finally {
--- a/src/main/java/io/cryostat/agent/remote/EventTemplatesContext.java
+++ b/src/main/java/io/cryostat/agent/remote/EventTemplatesContext.java
@@ -53,7 +53,7 @@ class EventTemplatesContext implements RemoteContext {
             switch (mtd) {
                 case "GET":
                     try {
-                        exchange.sendResponseHeaders(HttpStatus.SC_OK, 0);
+                        exchange.sendResponseHeaders(HttpStatus.SC_OK, BODY_LENGTH_UNKNOWN);
                         try (OutputStream response = exchange.getResponseBody()) {
                             FlightRecorderMXBean bean =
                                     ManagementFactory.getPlatformMXBean(FlightRecorderMXBean.class);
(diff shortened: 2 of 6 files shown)


src/main/java/io/cryostat/agent/remote/EventTemplatesContext.java
```diff
@@ -43,7 +43,7 @@ class EventTemplatesContext implements RemoteContext {
 
     @Override
     public String path() {
-        return "/event-templates";
+        return "/event-templates/";
```
mwangggg, line 46: Somehow this is causing the HTTP connection to close, which triggers an exception when trying to create a recording for the agent

mwangggg: Only "ALL" event template is available when trying to create a recording

andrewazores: I believe the second part about only "ALL" being available happens because of how the JDK HTTP server handles request path resolution. `/event-templates` will match any path prefixed with that string (ex. it would match `/event-templates` but also `/event-templates-and-more` or `/event-templates/and/more`), whereas `/event-templates/` will require the request to be exactly `/event-templates/`.

The Cryostat server would previously send a request on `/event-templates`, but in this PR:

https://github.com/cryostatio/cryostat/pull/1604/files#diff-2241c6f25048112e06962b86cb0f144af726370b545c86c3e9b0e49e0fe29bb5R272

I updated it so that the server requests `/event-templates/`.

If there is a mismatch where the Agent API exposes/expects `/event-templates/` and the Cryostat server asks it for `/event-templates`, I think the Agent responds to the server with an HTTP 404, which the server handles by logging an exception and using an empty list as the response for the available remote target templates. That empty list gets merged with any custom templates on the server, plus the `"ALL"` meta-template.

mwangggg: oops sorry I'm silly and didn't see the corresponding cryostat PR

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -163,7 +163,8 @@ private HttpHandler wrap(HttpHandler handler) {
                 handler.handle(x);
             } catch (Exception e) {
                 log.error("Unhandled exception", e);
-                x.sendResponseHeaders(HttpStatus.SC_INTERNAL_SERVER_ERROR, 0);
+                x.sendResponseHeaders(
+                        HttpStatus.SC_INTERNAL_SERVER_ERROR, RemoteContext.BODY_LENGTH_NONE);
```
mwangggg, line 167: should be RemoteContext.BODY_LENGTH_UNKNOWN

src/main/java/io/cryostat/agent/remote/EventTypesContext.java
```diff
@@ -59,17 +59,19 @@ public void handle(HttpExchange exchange) throws IOException {
                         events.addAll(getEventTypes());
                     } catch (Exception e) {
                         log.error("events serialization failure", e);
-                        exchange.sendResponseHeaders(HttpStatus.SC_INTERNAL_SERVER_ERROR, 0);
+                        exchange.sendResponseHeaders(
+                                HttpStatus.SC_INTERNAL_SERVER_ERROR, BODY_LENGTH_NONE);
```
mwangggg, line 63: BODY_LENGTH_UNKNOWN ?

src/main/java/io/cryostat/agent/WebServer.java
andrewazores: Hmm. Should it? If there is an unhandled internal exception then I think what should happen is the Agent responds to the requesting client with an HTTP 500 (that's the `SC_INTERNAL_SERVER_ERROR`), and this response would only include headers. The response body is empty since there is no content to report when an exception occurred.

I guess it's possible that the endpoint context handler has already started writing a response to the context stream, then hits an exception and here we try to send a "no body" header when the response is already partially written. I'm not sure exactly how the agent's HTTP server handles that, but the requesting client should hopefully take that as a sign to drop the stream if it starts receiving any response body.

mwangggg: ah I see - I thought you were just exchanging the numerical body lengths with constants

andrewazores: Mostly - I think these two cases were actually bugs previously (probably because the 0/-1 constants are not very readable), and the new constant represents the correct behaviour.

andrewazores: They are both references to the same constant, which form is used just depends on the lexical scope of the reference. Classes implementing the `RemoteContext` interface have the constant available in scope so they can use the bare `BODY_LENGTH_UNKNOWN`, whereas other classes like `WebServer` need to either `import i.c.a.r.RemoteContext` and use `RemoteContext.BODY_LENGTH_UNKNOWN`, or could `import static i.c.a.r.RemoteContext.BODY_LENGTH_UNKNOWN` and then use the bare reference as well.

mwangggg: Hmm what about MBeanContext that implements RemoteContext but uses RemoteContext.BODY_LENGTH_UNKNOWN ?

andrewazores: I can change that if you'd like, but it doesn't actually mean anything different.

mwangggg: I just figured for consistency sake it might be less confusing, but otherwise everything looks good

mwangggg (approved):

Issue #185, opened 2023-08-14
[Task] Accept request for creating JFR snapshots


andrewazores:

ishridharhegde:
Hi, since this is marked as **good first issue**, I'm interested to work on it. 
Can someone update the description with more info on what needs to be done?

andrewazores:
Finished in #186

Issue #162, opened 2023-06-28
[Story] Dynamically start Flight Recordings


andrewazores:
- [x] #163
- [x] #164
- [x] #185

andrewazores:
Done

PR #186, opened 2023-08-16
feat(api): enable JFR Snapshot

mwangggg:
fixes: #185 
fixes: #162
Related to: #124


src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -193,13 +192,26 @@ private void handleGetRecording(HttpExchange exchange, long id) {
...
+                    SerializableRecordingDescriptor snapshot = startSnapshot(req, exchange);
+                    exchange.sendResponseHeaders(HttpStatus.SC_CREATED, BODY_LENGTH_UNKNOWN);
+                    try (OutputStream response = exchange.getResponseBody()) {
+                        mapper.writeValue(response, snapshot);
+                    }
+                    return;
+                } catch (IOException e) {
+                    log.error("Failed to start snapshot", e);
+                    exchange.sendResponseHeaders(HttpStatus.SC_SERVICE_UNAVAILABLE, BODY_LENGTH_NONE);
+                }
```
andrewazores, line 213: the `return` statement should either be at the end here, or the code lower down that calls into `startRecording` should be in an `else` so that this doesn't accidentally handle an IOException, send headers, and then continue to try to start a recording with an invalid request body.

andrewazores:
Looks like a bad rebase or something. `git fetch upstream && git rebase -i upstream/124-epic-two-way-communications-protocol`, drop all commits that aren't ones you authored directly for this PR, and force-push back onto your branch.

mwangggg:
not really sure what happened there... my laptop has been bugging all morning

andrewazores:
`git rebase main` would have done something like that, I think. Easy mistake to make when working with upstream feature branches like this. Ideally I would have been keeping that one up to date with `main`, but I'll have to get around to it after this is merged.

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -346,12 +375,20 @@ boolean requestsBundledTemplate() {
...
+            return snapshotName && snapshotTemplate && snapshotFeatures;
+        }
+
         boolean isValid() {
             boolean requestsCustomTemplate = requestsCustomTemplate();
             boolean requestsBundledTemplate = requestsBundledTemplate();
             boolean requestsEither = requestsCustomTemplate || requestsBundledTemplate;
             boolean requestsBoth = requestsCustomTemplate && requestsBundledTemplate;
-            return requestsEither && !requestsBoth;
+            return requestsEither && !requestsBoth && !requestSnapshot();
```
andrewazores, line 391: I think this might be more understandable written as `requestSnapshot || (requestsEither && !requestsBoth /* ie requests exactly one */)`

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -106,12 +108,23 @@ public void handle(HttpExchange exchange) throws IOException {
...
-                    if (id >= 0) {
-                        handleStop(exchange, id);
+                    String request = requestStopOrUpdate(exchange);
+                    if (request == "STOPPED") {
+                        if (id >= 0) {
+                            handleStop(exchange, id);
+                        } else {
+                            exchange.sendResponseHeaders(HttpStatus.SC_BAD_REQUEST, BODY_LENGTH_NONE);
+                        }
+                    } else if (request == "UPDATE") {
```
andrewazores, line 123: The values for `request` should only be the [`RecordingState`s](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/RecordingState.html). I think you should be able to refactor this so that the Agent accepts the request, parses the body as JSON, iterates over each key in the JSON to determine which property should be set, then validates that the corresponding JSON value is acceptable (ex. for `state` that it can be mapped from the `RecordingState`), and finally performs some update operation on the local recording. The request state check shouldn't be done as a separate top-level check like this, in other words, it should be treated the same as the other properties.

andrewazores: This would mean it is possible for a client to issue a request that both renames and stops a recording simultaneously, for example, which would be fine, but your current implementation would not allow.

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -106,12 +109,23 @@ public void handle(HttpExchange exchange) throws IOException {
...
                 case "POST":
-                    handleStart(exchange);
+                    handleStartRecordingOrSnapshot(exchange);
                     break;
                 case "PATCH":
                     id = extractId(exchange);
-                    if (id >= 0) {
-                        handleStop(exchange, id);
+                    String request = requestStopOrUpdate(exchange);
+                    if (request == "STOPPED") {
```
andrewazores, line 117: Careful - in Java, Strings should be compared like `x.equals("A")`, not `x == "A"`, since the `==` only checks reference equality.

andrewazores:
Rebase onto `upstream/124-epic-two-way-communication`, not `main`

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -217,22 +236,81 @@ private void handleStart(HttpExchange exchange) throws IOException {
...
+                        break;
+                    default:
+                        log.warn("Unknown recording option {}", field.getKey());
+                        exchange.sendResponseHeaders(
+                                HttpStatus.SC_METHOD_NOT_ALLOWED, BODY_LENGTH_NONE);
+                        break;
+                }
+            }
+            svc.updateRecordingOptions(dsc, builder.build());
+            exchange.sendResponseHeaders(HttpStatus.SC_CREATED, BODY_LENGTH_UNKNOWN);
```
andrewazores, line 287: I would expect `CREATED` to be the successful response to a `POST`, and it would normally include a `Location` header as well.

Here I would use `200 OK` and `BODY_LENGTH_UNKNOWN` if the context is going to write back the updated recording descriptor as the response. If there is no response body then it would be `204 NO CONTENT` and `BODY_LENGTH_NONE`.

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -217,22 +236,81 @@ private void handleStart(HttpExchange exchange) throws IOException {
...
+
+            while (fields.hasNext()) {
+                Map.Entry<String, JsonNode> field = fields.next();
+
+                switch(field.getKey()){
+                    case "state":
+                        if ("STOPPED".equals(field.getValue().toString())) {
+                            handleStop(exchange, id);
+                            break;
+                        } else if (!"UPDATE".equals(field.getValue().toString())) {
```
andrewazores, line 260: Can this `state: "UPDATE"` case be removed? I think the only time the client needs to specify the `state` key currently is if they are trying to put the recording into `STOPPED` state - so if the value is anything other than `STOPPED`, it should result in a `BAD REQUEST`. If the client just wants to perform an update of other recording options without changing the state, they simply don't pass the `state` in the request.

src/main/java/io/cryostat/agent/remote/RecordingsContext.java
```diff
@@ -217,22 +236,81 @@ private void handleStart(HttpExchange exchange) throws IOException {
...
+                            exchange.sendResponseHeaders(HttpStatus.SC_BAD_REQUEST, BODY_LENGTH_NONE);
+                        }
+                        break;
+                    case "name":
+                        builder = builder.name(field.getValue().toString());
+                        break;
+                    case "duration":
+                        builder = builder.duration(field.getValue().asLong());
```
andrewazores, line 268: If one of these fails, ex. `getValue().asLong()` fails because the client provided a non-`Long`-parseable value, I assume this throws an Exception of some kind? Do we map that out to a `BAD REQUEST` somewhere?

mwangggg: no- I'll add that in

mwangggg: according to [this](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/JsonNode.html#asLong(long)) no exception is thrown and a default value is returned, which is not the behaviour we want, so I'll figure out a different way to convert the type

andrewazores (approved):

PR #195, opened 2023-09-18
fix(mbean): suppress warning message if BootClassPath unsupported

andrewazores:
Fixes https://github.com/cryostatio/cryostat-core/issues/218


aali309 (approved):
Looks good to me.

PR #203, opened 2023-09-20
feat(discovery): dual HTTP+JMX registration

### LINKED ISSUE #166: [Story] Agent dual-registration
Currently, the Agent implements a Cryostat Discovery Plugin by creating a plugin registration (and corresponding `REALM` node), and defines a single target JVM node within that Realm, representing itself. This will normally have an Agent HTTP `connectUrl`, unless the config property for `prefer-jmx` is set, in which case the Agent will publish a JMX ServiceURL for itself if the JVM it's attached to appears to have JMX enabled.

Why not publish both URLs? The Agent HTTP URL will always be available and publish-able as one target JVM node. After #163 this URL may also support write/mutation operations. A simple check endpoint can be added to return that status so that the available features can be queried without first trying to perform a write operation and seeing if it fails. A second target JVM node under the same Realm could be optionally published as well, if the host JVM appears to have JMX enabled. Then the `prefer-jmx` config can be dropped.

andrewazores:
Fixes #166

The Agent will always register as a discovery plugin and publish at least one JVM node describing itself as an Agent HTTP connection. If the Agent detects that JMX is also available on the target JVM then it will also publish a second JVM node describing itself with that JMX URL for connection.

image


aali309 (commented):
Looks good to me.

aali309 (approved):

PR #214, opened 2023-10-02
fix(logging): use slf4j-simple to avoid conflicts

### LINKED ISSUE #213: [Bug] `pom.xml` uses slf4j-jdk14
There is a [user report](https://groups.google.com/g/cryostat-development/c/PIUVtTrtfJs/m/U9nrk_N-AwAJ) that the Agent's use of `slf4j-jdk14` causes a StackOverflowError because of some configuration of the Spring Boot application that the Agent is attached to causing log calls to recurse into themselves. There is probably a smarter long-term fix, but in the short term for the next Agent release this could be worked around by using the `slf4j-simple` provider which simply echoes to stderr.

andrewazores:
Fixes #213
See also https://github.com/cryostatio/cryostat-core/pull/268

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,7 @@
   </dependency>
   <dependency>
     <groupId>org.slf4j</groupId>
-    <artifactId>slf4j-jdk14</artifactId>
+    <artifactId>slf4j-simple</artifactId>
     <version>${org.slf4j.version}</version>
   </dependency>
 


ebaron (approved):

andrewazores:
This will probably be obsoleted by https://github.com/cryostatio/cryostat-agent/pull/224

Issue #138, opened 2023-06-06
[Story] Separate TLS keystore/truststore from host application


andrewazores:
The Agent contains an HTTP webserver that exposes its API for Cryostat. This webserver should support HTTPS, and it should be possible to configure the Agent to use a TLS keystore and certificate for securing this HTTPS server. It should be possible for this TLS keystore to be separate from any the attached target application might already have.

The Agent also contains an HTTP client that it uses for communicating with Cryostat, to register itself as a discovery plugin, publsh information about itself, etc. This HTTP client should support HTTPS that the Cryostat server may expose. Likewise, it should be possible to configure the Agent to add the Cryostat server's TLS certificate to the Agent's truststore, and it should be possible to configure it so that this Agent truststore is separate from the truststore of the attached target application.

- [x] #139
- [x] #140
- [x] #141
- [x] #142
- [x] https://github.com/cryostatio/cryostat-agent/issues/460

Related: https://github.com/cryostatio/cryostat-operator/issues/595

andrewazores:
Another possibility: generate a certificate in advance, common to all Agent instances. The certificate's domain will be incorrect in this case so the Cryostat server's webclient will need to ignore host verification. This certificate would be bundled as a resource in the Agent JAR as well as included in the Cryostat server's default truststore. This way, the Agent instances can always present an HTTPS API and all data leaving the Agent can always be encrypted with this cert, and the Cryostat server can at least trust this certificate for its Agent webclients, even if it cannot verify the hostname. This way there is a zero-configuration setup that allows HTTPS to be always on for Cryostat<->Agent communications.

In the future this could be extended with a configurable setup that allows the user to supply their own cert as originally outlined at the top of this issue. The Operator could be in charge of supplying this cert to the Agent for it to use for its embedded HTTPS server, and to the Cryostat instance in its truststore. This would allow more granularity so that the certificate is common to all Agents within a given Namespace, rather than all Agent instances indiscriminately.

andrewazores:
https://stackoverflow.com/questions/2308479/simple-java-https-server

andrewazores:
I think this is actually already completed with @mwangggg's latest work on the Agent truststore/keystore, where those are already independent stores from the global JVM configuration.

PR #257, opened 2023-11-21
feat(HTTPS): implement HTTPS keystore

mwangggg:
fixes: #141
Related to: https://github.com/cryostatio/cryostat-agent/issues/138#issuecomment-1728040739


src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -82,39 +102,111 @@ class WebServer {
...
-                        rc -> {
-                            HttpContext ctx = this.http.createContext(rc.path(), wrap(rc::handle));
-                            ctx.setAuthenticator(agentAuthenticator);
-                            ctx.getFilters().add(requestLoggingFilter);
-                            ctx.getFilters().add(compressionFilter);
-                        });
+        try {
+            // initialize new HTTPS server
+            this.https = HttpsServer.create(new InetSocketAddress(host, port), 0);
+            SSLContext sslContext = SSLContext.getInstance("TLS");
```
andrewazores, line 120: I think this should be `getInstance("TLSv1.2")` or `TLSv1.3`. Same in `MainModule`.

andrewazores: https://github.com/cryostatio/cryostat-agent/issues/343

src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -4,23 +4,20 @@ cryostat.agent.baseuri=
 
 cryostat.agent.api.writes-enabled=false
 
-# TODO rename these *.ssl.* properties to *.tls.* , and update the README for all these properties
-cryostat.agent.webclient.ssl.trust-all=false
-cryostat.agent.webclient.ssl.verify-hostname=true
+cryostat.agent.webclient.tls.version=TLSv1.2
```
andrewazores, line 7: Add some documentation to the README for this and the other new properties, please

README.md
```diff
@@ -139,17 +139,26 @@ and how it advertises itself to a Cryostat server instance. Required properties
...
 - [ ] `cryostat.agent.realm` [`String`]: the Cryostat Discovery API "realm" that this agent belongs to. This should be unique per agent instance. The default is the value of `cryostat.agent.app.name`.
-- [ ] `cryostat.agent.authorization` [`String`]: Authorization header value to include with API requests to the Cryostat server, ex. `Bearer abcd1234`. Default `None`.
-- [ ] `cryostat.agent.webclient.ssl.trust-all` [`boolean`]: Control whether the agent trusts all certificates presented by the Cryostat server. Default `false`. This should only be overridden for development and testing purposes, never in production.
-- [ ] `cryostat.agent.webclient.ssl.verify-hostname` [`boolean`]: Control whether the agent verifies hostnames on certificates presented by the Cryostat server. Default `true`. This should only be overridden for development and testing purposes, never in production.
+- [ ] `cryostat.agent.authorization` [`String`]: authorization header value to include with API requests to the Cryostat server, ex. `Bearer abcd1234`. Default `None`.
```
andrewazores, line 146: I think `Authorization` can remain capitalized here, since conventionally HTTP headers are written in a `Dashed-Title-Case` format. Maybe surrounding it in backticks for monospace formatting would help make that point that it's referring specifically to that header, not just the notion of authorization.

README.md
```diff
@@ -139,17 +139,26 @@ and how it advertises itself to a Cryostat server instance. Required properties
...
+- [ ] `cryostat.agent.webclient.tls.verify-hostname` [`boolean`]: control whether the agent verifies hostnames on certificates presented by the Cryostat server. Default `true`. This should only be overridden for development and testing purposes, never in production.
 - [ ] `cryostat.agent.webclient.connect.timeout-ms` [`long`]: the duration in milliseconds to wait for HTTP requests to the Cryostat server to connect. Default `1000`.
 - [ ] `cryostat.agent.webclient.response.timeout-ms` [`long`]: the duration in milliseconds to wait for HTTP requests to the Cryostat server to respond. Default `1000`.
 - [ ] `cryostat.agent.webserver.host` [`String`]: the internal hostname or IP address for the embedded webserver to bind to. Default `0.0.0.0`.
 - [ ] `cryostat.agent.webserver.port` [`int`]: the internal port number for the embedded webserver to bind to. Default `9977`.
+- [ ] `cryostat.agent.webserver.tls.version` [`String`]: the version of TLS used for the Agent's server SSL context. Default `TLSv1.2`.
+- [ ] `cryostat.agent.webserver.tls.keystore.pass` [`String`]: the name of the file containing the HTTPS server keystore's password
```
andrewazores, line 155: This is a file path, isn't it? Not just a filename

README.md
```diff
@@ -139,17 +139,26 @@ and how it advertises itself to a Cryostat server instance. Required properties
...
+- [ ] `cryostat.agent.webserver.tls.keystore.pass.charset` [`String`]: the character set used by the HTTPS server keystore's password. Default `utf-8`.
+- [ ] `cryostat.agent.webserver.tls.keystore.file` [`String`]: the file path to the HTTPS server keystore
+- [ ] `cryostat.agent.webserver.tls.keystore.type` [`String`]: the type of keystore used for the Agent's HTTPS server. Default `PKCS12`.
+- [ ] `cryostat.agent.webserver.tls.cert.alias` [`String`]: the alias for the certificate stored in the HTTPS server keystore. Default `serverCert`.
+- [ ] `cryostat.agent.webserver.tls.cert.file` [`String`]: the name of the file containing the certificate to be stored by the HTTPS server keystore
```
andrewazores, line 160: Same here, file path instead of file name?

src/main/resources/certs/generate-dev-certs.sh
andrewazores, line 1: I think this should be moved out of `src/main` and just into the top-level of the project directory, so that it doesn't end up packaged into the JARs for distribution.

src/main/resources/certs/cryostat-keystore.p12
andrewazores, line 1: This was just for development/testing, right? I think it should be removed, along with the `server.cer`. If we want to have some self-signed certificate available by default then I think this would be something we generate during CI at publish time, not checked in to source control.

pom.xml
```diff
@@ -269,6 +269,18 @@
...
+          <exclude>**/*.cer</exclude>
+      </excludes>
+    </resource>
+    <resource>
+      <directory>src/main/resources</directory>
+      <filtering>false</filtering>
+      <includes>
+          <include>**/*.p12</include>
+          <include>**/*.cer</include>
+      </includes>
```
andrewazores, line 283: These changes can be removed now, I think

andrewazores:
Change looks good but I will wait to approve/merge until we have demonstrated that it works as expected with a test case where the certificate is supplied by mounting it to an application container image containing the Agent.

certs/generate-dev-certs.sh
andrewazores, line 1: I don't think we necessarily need this anymore, since something equivalent is included in the Cryostat server-side PR for smoketesting, right?

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -152,13 +173,92 @@ public X509Certificate[] getAcceptedIssuers() {
...
+                    KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+            kmf.init(ks, password.toCharArray());
+
+            // set up trust manager factory
+            TrustManagerFactory tmf =
+                    TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            tmf.init(ks);
+
+            // set up HTTPS context
+            sslContext.init(kmf.getKeyManagers(), tmf.getTrustManagers(), null);
```
andrewazores, line 242: On line 169 when we set up the SSLContext for the Agent's HTTP Client, we supply a `new SecureRandom()` here. I think this ends up being the same thing in effect, since it looks like the JDK library code will also just call `new SecureRandom()` if `null` is passed, so maybe that other line needs to be adjusted to match this one.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -186,6 +286,46 @@ public static HttpClient provideHttpClient(
...
+    public static HttpServer provideHttpServer(
+            ScheduledExecutorService executor,
+            @Named(ConfigModule.CRYOSTAT_AGENT_WEBSERVER_HOST) String host,
+            @Named(ConfigModule.CRYOSTAT_AGENT_WEBSERVER_PORT) int port,
+            @Named(HTTP_SERVER_SSL_CTX) Optional<SSLContext> sslContext) {
+        try {
+            HttpServer http;
+            if (sslContext.isEmpty()) {
+                http = HttpServer.create(new InetSocketAddress(host, port), 0);
+                http.setExecutor(executor);
```
andrewazores, line 300: `setExecutor` is also called down on line 321

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -52,14 +51,12 @@ class WebServer {
...
     private final Lazy<CryostatClient> cryostat;
-    private final ScheduledExecutorService executor;
-    private final String host;
-    private final int port;
     private final Credentials credentials;
     private final URI callback;
     private final Lazy<Registration> registration;
     private HttpServer http;
     private volatile int credentialId = -1;
+    private final FileSystem fs;
```
andrewazores, line 59: This appears to be unused

ebaron (commented):
This seems to cover the server TLS config. Is there something similar planned for the client?

mwangggg:
> This seems to cover the server TLS config. Is there something similar planned for the client?

I'm currently working on the client truststore, but I'll open a new PR for that

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -263,6 +389,12 @@ public static FlightRecorderHelper provideFlightRecorderHelper() {
         return new FlightRecorderHelper();
     }
 
+    @Provides
+    @Singleton
+    public static FileSystem provideFileSystem() {
```
andrewazores, line 394: I think this is also unneeded now since the injection point was removed?

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -186,6 +276,43 @@ public static HttpClient provideHttpClient(
...
+                                    params.setNeedClientAuth(false);
+                                    params.setCipherSuites(engine.getEnabledCipherSuites());
+                                    params.setProtocols((engine.getEnabledProtocols()));
+                                    params.setSSLParameters(context.getDefaultSSLParameters());
+                                } catch (Exception e) {
+                                    throw new RuntimeException(e);
+                                }
+                            }
+                        });
+                http = https;
```
andrewazores, line 307:
```suggestion
                http = HttpsServer.create(new InetSocketAddress(host, port), 0);
                ((HttpsServer) http).setHttpsConfigurator(
                        new HttpsConfigurator(sslContext.get()) {
                            public void configure(HttpsParameters params) {
                                try {
                                    SSLContext context = getSSLContext();
                                    SSLEngine engine = context.createSSLEngine();
                                    params.setNeedClientAuth(false);
                                    params.setCipherSuites(engine.getEnabledCipherSuites());
                                    params.setProtocols((engine.getEnabledProtocols()));
                                    params.setSSLParameters(context.getDefaultSSLParameters());
                                } catch (Exception e) {
                                    throw new RuntimeException(e);
                                }
                            }
                        });
```

Not sure if I really like this any better after having written it out. Feel free to take it or leave it.

andrewazores:
Just needs a `spotless` run, looks good otherwise.

andrewazores (approved):

PR #263, opened 2023-11-24
build(jackson): use jackson-bom

andrewazores:


andrewazores:
I'll hold this until 2.4.0 is out the door since that is just around the corner, and merge this right after so it can go into 2.4.1.

PR #276, opened 2023-11-29
test(smart_triggers): add test case for single-quoted duration

andrewazores:
Related to #253

Adds one extra test case to ensure that the `duration()` syntax accepts single-quoted values. This can be useful when deploying the Agent so the user has flexibility on where to use single and double quotes to avoid having to use escape sequences while still being able to use whitespace, or perhaps even variable interpolations, etc.

The implementation already handles this case so this is simply adding a test to catch any future regressions.


mwangggg (approved):
lgtm

Issue #291, opened 2023-12-05
[Bug] Agent cannot accept `ALL` event template in recording start requests


andrewazores:
![image](https://github.com/cryostatio/cryostat-agent/assets/3787464/94e78b5e-12ef-4d6f-a848-13d9b9266ac0)

Target event templates and custom event templates are properly handled and implemented between the Cryostat server and the Agent, but the `ALL` meta-template is not understood by the Agent.

andrewazores:
@aali309 @mwangggg either of you interested in picking this up?

PR #310, opened 2023-12-22
fix(agent): accept 'ALL' event template in recording start

aali309:
Fixes: #291

### DIFF
--- a/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
+++ b/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
@@ -75,7 +75,9 @@ public class FlightRecorderHelper {
             Recording recording = new Recording(allTemplate);
             recording.setToDisk(true);
 
-            return Optional.of(new TemplatedRecording(null, recording));
+            return Optional.of(
+                    new TemplatedRecording(
+                            new ConfigurationInfo(TemplateType.TARGET, "ALL", "ALL"), recording));
         } else {
             Optional<Configuration> opt = getTemplate(templateNameOrLabel);
             if (opt.isEmpty()) {
@@ -89,7 +91,8 @@ public class FlightRecorderHelper {
             Recording recording = new Recording(configuration.getSettings());
             recording.setToDisk(true);
 
-            return Optional.of(new TemplatedRecording(configuration, recording));
+            return Optional.of(
+                    new TemplatedRecording(new ConfigurationInfo(configuration), recording));
         }
     }
 
@@ -125,21 +128,55 @@ public class FlightRecorderHelper {
 
     @SuppressFBWarnings(value = {"EI_EXPOSE_REP", "EI_EXPOSE_REP2"})
     public static class TemplatedRecording {
-        private final Configuration configuration;
+        private final ConfigurationInfo configuration;
         private final Recording recording;
 
-        public TemplatedRecording(Configuration configuration, Recording recording) {
-            this.configuration = configuration;
-            this.recording = recording;
+        public TemplatedRecording(ConfigurationInfo configuration, Recording recording) {
+            this.configuration = Objects.requireNonNull(configuration);
+            this.recording = Objects.requireNonNull(recording);
         }
 
-        public Configuration getConfiguration() {
+        public ConfigurationInfo getConfigurationInfo() {
             return configuration;
         }
 
         public Recording getRecording() {
             return recording;
         }
+    }
+
+    public static class ConfigurationInfo {
+        private final String name;
+        private final String label;
+        private final TemplateType type;
+
+        public ConfigurationInfo(TemplateType type, String name, String label) {
+            this.type = Objects.requireNonNull(type);
+            this.name = Objects.requireNonNull(name);
+            this.label = Objects.requireNonNull(label);
+        }
+
+        public ConfigurationInfo(Configuration configuration) {
+            this(TemplateType.TARGET, configuration.getName(), configuration.getLabel());
+        }
+
+        public String getName() {
+            return name;
+        }
+
+        public String getLabel() {
+            return label;
+        }
+
+        public TemplateType getType() {
+            return type;
+        }
+    }
+
+    public enum TemplateType {
+        TARGET,
+        CUSTOM,
+        ;
     }
 
     @SuppressFBWarnings(value = "URF_UNREAD_FIELD")
--- a/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
+++ b/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
@@ -18,6 +18,7 @@ package io.cryostat.agent;
 import java.io.IOException;
 import java.io.StringReader;
 import java.text.ParseException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -55,17 +56,43 @@ public class FlightRecorderHelper {
 
     public Optional<TemplatedRecording> createRecordingWithPredefinedTemplate(
             String templateNameOrLabel) {
-        Optional<Configuration> opt = getTemplate(templateNameOrLabel);
-        if (opt.isEmpty()) {
-            log.error(
-                    "Cannot start recording with template named or labelled {}",
-                    templateNameOrLabel);
-            return Optional.empty();
+        if ("ALL".equals(templateNameOrLabel)) {
+            Map<String, String> allTemplate = new HashMap<>();
+            FlightRecorder.getFlightRecorder()
+                    .getEventTypes()
+                    .forEach(
+                            e -> {
+                                e.getSettingDescriptors()
+                                        .forEach(
+                                                s -> {
+                                                    allTemplate.put(
+                                                            e.getName() + "#" + s.getName(),
+                                                            s.getDefaultValue());
+                                                });
+                                allTemplate.put(e.getName() + "#enabled", "true");
+                            });
+
+            Recording recording = new Recording(allTemplate);
+            recording.setToDisk(true);
+
+            log.info("Creating recording with ALL template");
+            return Optional.of(new TemplatedRecording(null, recording));
+        } else {
+            Optional<Configuration> opt = getTemplate(templateNameOrLabel);
+            if (opt.isEmpty()) {
+                log.error(
+                        "Cannot start recording with template named or labelled {}",
+                        templateNameOrLabel);
+                return Optional.empty();
+            }
+
+            Configuration configuration = opt.get();
+            Recording recording = new Recording(configuration.getSettings());
+            recording.setToDisk(true);
+
+            log.info("Creating recording with template: {}", templateNameOrLabel);
+            return Optional.of(new TemplatedRecording(configuration, recording));
         }
-        Configuration configuration = opt.get();
-        Recording recording = new Recording(configuration.getSettings());
-        recording.setToDisk(true);
-        return Optional.of(new TemplatedRecording(configuration, recording));
     }
 
     public Optional<Configuration> getTemplate(String nameOrLabel) {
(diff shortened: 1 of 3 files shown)


aali309:
Agent Logs:
```2023-12-22 17:50:49:010 +0000 [cryostat-agent-worker-1] INFO io.cryostat.agent.WebServer - Using 'deflate' encoding
2023-12-22 17:50:49:017 +0000 [cryostat-agent-worker-1] INFO io.cryostat.agent.FlightRecorderHelper - Creating recording with ALL template
2023-12-22 17:50:49:090 +0000 [cryostat-agent-worker-1] INFO io.cryostat.agent.harvest.Harvester - test2(2) RUNNING
2023-12-22 17:50:49:091 +0000 [cryostat-agent-worker-1] INFO io.cryostat.agent.WebServer - POST /recordings/ : 201 81ms
2023-12-22 17:50:49:270 +0000 [cryostat-agent-worker-0] INFO io.cryostat.agent.WebServer - GET /recordings/
2023-12-22 17:50:49:270 +0000 [cryostat-agent-worker-0] INFO io.cryostat.agent.WebServer - Using 'deflate' encoding
2023-12-22 17:50:49:271 +0000 [cryostat-agent-worker-0] INFO io.cryostat.agent.WebServer - GET /recordings/ : 200 1ms
2023-12-22 17:50:53:412 +0000 [cryostat-agent-registration] INFO io.cryostat.agent.CryostatClient - GET https://localhost:8181/api/v2.2/discovery/ca380842-a4c2-43a0-804f-c9820690d82e?token=eyJjdHkiOiJKV1QiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..QEdCYH18em4C2pbs.kJCntPnvp8ZbJlP0HUG--18zxZSnM0DwgH7u9AYXCA7lnzyaJBMEeqLLYiNrreEqDsobHTVlNXQuRd6YwAAAMsrZ5Ugqpke7F9Yy3-urddLAA_OTj0uSp7rUelzBl16UTi39gZoIuDIqeHIPxjgDQO8hTspPsXd8MvNrYa8n8nHNlOtxqtUDgqrWJdwoCeQamD6uZKQpHsRPRHmxz3uujOAbPK2zg20l00HMRVM_Q6sPzYodd8S8QIbvSPQZvxJaG7og9QPOT63BJ7GbudQngm-vuOiWw7J_JKlKUP9lI9rWin6ikwGxS9-sT44Au1Q-_dw7jv_sZmfd-w6Skt1pA49Fcr0Q0U4EuOBSpbvnRQIV38hzxEDmUcC4Hjp4XKzwopUAfRLI1Bqm5whQcCWzmK2PCofieMmPyPZWwn7RhrLYvy6sT9FYopwHgW7XqmN07q-pX9kDc98iny_Qr4uVjcFs5RgfupN7cuhHPLl1LdEUnBCKZe9IbfZW-lxUlgYXOAPShyKJ_tI8xPeExQ_fMW8YiUC1XAhupYlSD1mYkQSMNISMCwBqjT75wMMXXrYtXt5qyRsQGbUu.qNOIukMIgT_nHKZ08hzJiA HTTP/1.1
```
Screenshot from 2023-12-22 12-51-20

Screenshot from 2023-12-22 12-55-21

src/main/java/io/cryostat/agent/FlightRecorderHelper.java
```diff
@@ -55,17 +56,41 @@ public Recording createRecordingWithCustomTemplate(String template)
...
+                                                            e.getName() + "#" + s.getName(),
+                                                            s.getDefaultValue());
+                                                });
+                                allTemplate.put(e.getName() + "#enabled", "true");
+                            });
+
+            Recording recording = new Recording(allTemplate);
+            recording.setToDisk(true);
+
+            return Optional.of(new TemplatedRecording(null, recording));
```
andrewazores, line 78: I don't think this `null` is safe. There is only one place that calls `templatedRecording.getConfiguration()`, and it assumes that it will be non-null:

https://github.com/cryostatio/cryostat-agent/blob/c28859afd1a5d8b210b52dfac3dba370c26880b8/src/main/java/io/cryostat/agent/CryostatClient.java#L376

Since this is the only accessor of the `configuration`, and all it actually wants from it is the name, I think we can do some light refactoring:

```patch
diff --git a/src/main/java/io/cryostat/agent/CryostatClient.java b/src/main/java/io/cryostat/agent/CryostatClient.java
index a6d8719..9a77e9c 100644
--- a/src/main/java/io/cryostat/agent/CryostatClient.java
+++ b/src/main/java/io/cryostat/agent/CryostatClient.java
@@ -37,6 +37,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.function.Function;
 
+import io.cryostat.agent.FlightRecorderHelper.ConfigurationInfo;
 import io.cryostat.agent.FlightRecorderHelper.TemplatedRecording;
 import io.cryostat.agent.WebServer.Credentials;
 import io.cryostat.agent.harvest.Harvester;
@@ -49,7 +50,6 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import jdk.jfr.Configuration;
 import jdk.jfr.Recording;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.input.CountingInputStream;
@@ -373,8 +373,8 @@ public class CryostatClient {
         Instant start = Instant.now();
         String timestamp = start.truncatedTo(ChronoUnit.SECONDS).toString().replaceAll("[-:]", "");
         String template =
-                opt.map(TemplatedRecording::getConfiguration)
-                        .map(Configuration::getName)
+                opt.map(TemplatedRecording::getConfigurationInfo)
+                        .map(ConfigurationInfo::getName)
                         .map(String::toLowerCase)
                         .map(String::trim)
                         .orElse("unknown");
diff --git a/src/main/java/io/cryostat/agent/FlightRecorderHelper.java b/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
index 65d9217..82a5058 100644
--- a/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
+++ b/src/main/java/io/cryostat/agent/FlightRecorderHelper.java
@@ -89,7 +89,8 @@ public class FlightRecorderHelper {
             Recording recording = new Recording(configuration.getSettings());
             recording.setToDisk(true);
 
-            return Optional.of(new TemplatedRecording(configuration, recording));
+            return Optional.of(
+                    new TemplatedRecording(new ConfigurationInfo(configuration), recording));
         }
     }
 
@@ -125,15 +126,15 @@ public class FlightRecorderHelper {
 
     @SuppressFBWarnings(value = {"EI_EXPOSE_REP", "EI_EXPOSE_REP2"})
     public static class TemplatedRecording {
-        private final Configuration configuration;
+        private final ConfigurationInfo configuration;
         private final Recording recording;
 
-        public TemplatedRecording(Configuration configuration, Recording recording) {
-            this.configuration = configuration;
-            this.recording = recording;
+        public TemplatedRecording(ConfigurationInfo configuration, Recording recording) {
+            this.configuration = Objects.requireNonNull(configuration);
+            this.recording = Objects.requireNonNull(recording);
         }
 
-        public Configuration getConfiguration() {
+        public ConfigurationInfo getConfigurationInfo() {
             return configuration;
         }
 
@@ -142,6 +143,40 @@ public class FlightRecorderHelper {
         }
     }
 
+    public static class ConfigurationInfo {
+        private final String name;
+        private final String label;
+        private final TemplateType type;
+
+        public ConfigurationInfo(TemplateType type, String name, String label) {
+            this.type = Objects.requireNonNull(type);
+            this.name = Objects.requireNonNull(name);
+            this.label = Objects.requireNonNull(label);
+        }
+
+        public ConfigurationInfo(Configuration configuration) {
+            this(TemplateType.TARGET, configuration.getName(), configuration.getLabel());
+        }
+
+        public String getName() {
+            return name;
+        }
+
+        public String getLabel() {
+            return label;
+        }
+
+        public TemplateType getType() {
+            return type;
+        }
+    }
+
+    public enum TemplateType {
+        TARGET,
+        CUSTOM,
+        ;
+    }
+
     @SuppressFBWarnings(value = "URF_UNREAD_FIELD")
     public static class RecordingInfo {
 

```

src/test/java/io/cryostat/agent/FlightRecorderHelperTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import io.cryostat.agent.FlightRecorderHelper.TemplatedRecording;
+
+import jdk.jfr.Recording;
+import org.junit.jupiter.api.Test;
+
+public class FlightRecorderHelperTest {
+    @Test
```
andrewazores, line 29: Could you make this a `@ParameterizedTest` and add cases for `default`, `Continuous`, `profile`, and `Profiling`? This way we can test that both of the standard OpenJDK built-in templates as well as our `ALL` "meta-template" all behave in the same way.

src/main/java/io/cryostat/agent/FlightRecorderHelper.java
```diff
@@ -75,7 +75,9 @@ public Optional<TemplatedRecording> createRecordingWithPredefinedTemplate(
             Recording recording = new Recording(allTemplate);
             recording.setToDisk(true);
 
-            return Optional.of(new TemplatedRecording(null, recording));
+            return Optional.of(
+                    new TemplatedRecording(
+                            new ConfigurationInfo(TemplateType.TARGET, "ALL", "ALL"), recording));
```
aali309, line 80: I think its better to have the `allTemplate` as the name.

andrewazores: Let's shelve that for a separate discussion. The convention has always been `ALL` here since the very early days of the project, so changing it or adding a new alias for it is actually a bigger question and should be handled separately from this PR I think.

andrewazores:
Looking good other than the comment thread above.

andrewazores (approved):

src/test/java/io/cryostat/agent/FlightRecorderHelperTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @ValueSource(strings = {"default", "Continuous", "profile", "Profiling", "ALL"})
+    public void testCreateRecordingWithTemplate(String template) {
+        FlightRecorderHelper helper = new FlightRecorderHelper();
+
+        Optional<TemplatedRecording> recording =
+                helper.createRecordingWithPredefinedTemplate(template);
+
+        switch (template) {
+            case "ALL":
+                assertNotNull(recording);
```
andrewazores, line 42: Actually, hang on - I think all of these cases should behave the same - the `Optional recording` should be not null and not empty for all of these `@ParameterizedTest` values.

src/test/java/io/cryostat/agent/FlightRecorderHelperTest.java
```diff
@@ -0,0 +1,89 @@
...
+                assertFalse(recording.isEmpty());
+                break;
+            case "profile":
+                assertFalse(recording.isEmpty());
+                break;
+            case "Profiling":
+                assertFalse(recording.isEmpty());
+                break;
+            default:
+                fail("Unexpected template: " + template);
```
andrewazores, line 58: It doesn't look like this `default` case ever gets hit, and if it did, wouldn't this `fail` cause a test failure before the `assertTrue`?

aali309: Yes, looks like the `default` case is not being hit.

src/test/java/io/cryostat/agent/FlightRecorderHelperTest.java
aali309: I think a better approach is to remove the switch statement and perform common assertions for all template `Label` or `name` values.

andrewazores (approved):

Issue #279, opened 2023-11-30
[Bug] Shaded JAR Manifest includes `Class-Path` entry


andrewazores:
The `Class-Path` entry is meant to list additional JARs that should be on the classpath along with the loaded JAR, but in this case we are talking about the shaded Agent JAR, so its dependencies are not loaded from the filesystem classpath but are rather bundled into the larger shaded JAR. It seems that the `Class-Path` manifest entry should simply be removed.

PR #311, opened 2023-12-24
fix(pom): remove classpath manifest entry

mwangggg:
fixes: #279 
I also replaced the deprecated OperatingSystemMXBean methods.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,7 +248,6 @@
       <configuration>
         <archive>
           <manifest>
-            <addClasspath>true</addClasspath>
             <mainClass>${mainClass}</mainClass>
           </manifest>
           <manifestEntries>


andrewazores:
> I also replaced the deprecated OperatingSystemMXBean methods.

Could you split these changes into a separate PR? I think we will want the classpath manifest piece to go into the next bugfix release, but the MXBean changes would need to wait for the next major or minor release.

andrewazores:
Actually, it looks like the MXBean changes would need to wait until we drop JDK  11 support:

```
 Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/cryostat-agent/cryostat-agent/src/main/java/io/cryostat/agent/model/MBeanInfo.java:[81,57] error: invalid method reference
  cannot find symbol
    symbol:   method getFreeMemorySize()
    location: interface UnixOperatingSystemMXBean
Error:  /home/runner/work/cryostat-agent/cryostat-agent/src/main/java/io/cryostat/agent/model/MBeanInfo.java:[85,48] error: invalid method reference
  cannot find symbol
    symbol:   method getCpuLoad()
    location: interface UnixOperatingSystemMXBean
Error:  /home/runner/work/cryostat-agent/cryostat-agent/src/main/java/io/cryostat/agent/model/MBeanInfo.java:[86,58] error: invalid method reference
  cannot find symbol
    symbol:   method getTotalMemorySize()
    location: interface UnixOperatingSystemMXBean
```

pom.xml
```diff
@@ -241,23 +241,6 @@
...
-      <artifactId>maven-jar-plugin</artifactId>
-      <version>${org.apache.maven.plugins.jar.version}</version>
-      <configuration>
-        <archive>
-          <manifest>
-            <addClasspath>true</addClasspath>
-            <mainClass>${mainClass}</mainClass>
-          </manifest>
-          <manifestEntries>
-            <Premain-Class>${mainClass}</Premain-Class>
```
andrewazores, line 255: I think we still need the premain and agent-class manifest attributes at least:

```
Starting the Java application using /opt/jboss/container/java/run/run-java.sh ...
INFO exec -a "java" java -XX:MaxRAMPercentage=80.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+ExitOnOutOfMemoryError -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -javaagent:/deployments/app/cryostat-agent.jar -cp "." -jar /deployments/quarkus-run.jar 
INFO running in /deployments
Error occurred during initialization of VM
agent library failed to init: instrument
Failed to find Premain-Class manifest attribute in /deployments/app/cryostat-agent.jar
```

andrewazores:
Looks good, at least when using the `-shaded` version of the Agent JAR. I haven't yet tested it with the non-shaded version but I suspect we might need to make this manifest attribute conditionally included only for that non-shaded form.

andrewazores:
Maybe we need to use a `-Pshaded` Maven profile for building shaded vs non-shaded. That way we can move this configuration into one configuration or the other so that the manifest gets built according to which way the JAR was built. @ebaron WDYT? Right now the build doesn't use any profiles, it just produces the plain, `-sources`, and `-shaded` JARs every time.

ebaron:
If there's some value in retaining the Class-Path attribute for non-shaded jars, then a profile could work. Although, even when building a shaded jar, the non-shaded jar is built too. We'd probably need to have it do both.

andrewazores:
I tried to put together a quick test of a sample application that consumes the non-shaded JAR, and it ended up with the application correctly loading the agent JAR but not having any of the agent dependencies on the classpath, so the agent failed miserably. That could be worked around at the application level by adding all of the agent's dependencies to the application classpath as well, but I think/hope that retaining the classpath manifest entry will avoid that step.

Then again, our primary usecase is for the shaded JAR, and I'm not sure how big of a hurdle it is for any users who do need the non-shaded JAR to also have to figure out the agent dependency management, too. If this ends up being too much work then maybe we just go without it for now and wait until someone actually asks for it.

andrewazores (approved):

PR #367, opened 2024-04-04
feat(webserver): password length and username can be configured

### LINKED ISSUE #341: [Task] Basic auth user should be configurable
https://github.com/cryostatio/cryostat-agent/blob/9647c68c2db8a53188a2127394bc758732342eed/src/main/java/io/cryostat/agent/WebServer.java#L298

andrewazores:
Fixes #341

- makes the webserver credential username configurable. Defaults to `user` as before, but this can be overridden if some need arises.
- makes the webserver credential password length configurable. This is always randomly generated, but the length was previously hardcoded to 24 characters. Now this can be configured, but continues to default to 24.
- expands the set of characters that can be generated for the password to include more of the ASCII symbols set. This increases password strength. Also simplifies the logic to just fill the password with any printable ASCII character, instead of trying to require at least one character from each of specific ranges first and then filling the remainder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,8 @@ and how it advertises itself to a Cryostat server instance. Required properties 
 - [ ] `cryostat.agent.webserver.host` [`String`]: the internal hostname or IP address for the embedded webserver to bind to. Default `0.0.0.0`.
 - [ ] `cryostat.agent.webserver.port` [`int`]: the internal port number for the embedded webserver to bind to. Default `9977`.
 - [ ] `cryostat.agent.webserver.credentials.user` [`String`]: the username used for `Basic` authorization on the embedded webserver. Default `user`.
-- [ ] `cryostat.agent.webserver.credentials.pass-length` [`int`]: the length of the generated password used for `Basic` authorization on the embedded webserver. Default `24`.
+- [ ] `cryostat.agent.webserver.credentials.pass.length` [`int`]: the length of the generated password used for `Basic` authorization on the embedded webserver. Default `24`.
+- [ ] `cryostat.agent.webserver.credentials.pass.hash-function` [`String`]: The name of the hash function to use when generating passwords. Default `SHA-256`.
 - [ ] `cryostat.agent.app.name` [`String`]: a human-friendly name for this application. Default `cryostat-agent`.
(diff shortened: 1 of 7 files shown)


src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -64,6 +66,12 @@ public abstract class ConfigModule {
 
     public static final String CRYOSTAT_AGENT_WEBSERVER_HOST = "cryostat.agent.webserver.host";
     public static final String CRYOSTAT_AGENT_WEBSERVER_PORT = "cryostat.agent.webserver.port";
+    public static final String CRYOSTAT_AGENT_WEBSERVER_CREDENTIALS_USER =
+            "cryostat.agent.webserver.credentials.user";
+    public static final String CRYOSTAT_AGENT_WEBSERVER_CREDENTIALS_PASS_HASH_FUNCTION =
```
mwangggg, line 71: I think this config property still needs to be added to the README

README.md
```diff
@@ -176,6 +176,9 @@ and how it advertises itself to a Cryostat server instance. Required properties
 - [ ] `cryostat.agent.webclient.response.timeout-ms` [`long`]: the duration in milliseconds to wait for HTTP requests to the Cryostat server to respond. Default `1000`.
 - [ ] `cryostat.agent.webserver.host` [`String`]: the internal hostname or IP address for the embedded webserver to bind to. Default `0.0.0.0`.
 - [ ] `cryostat.agent.webserver.port` [`int`]: the internal port number for the embedded webserver to bind to. Default `9977`.
+- [ ] `cryostat.agent.webserver.credentials.user` [`String`]: the username used for `Basic` authorization on the embedded webserver. Default `user`.
+- [ ] `cryostat.agent.webserver.credentials.pass.length` [`int`]: the length of the generated password used for `Basic` authorization on the embedded webserver. Default `24`.
+- [ ] `cryostat.agent.webserver.credentials.pass.hash-function` [`String`]: The name of the hash function to use when generating passwords. Default `SHA-256`.
```
mwangggg, line 181: lower case t in "the"

mwangggg (approved):
lgtm, just needs a rebase

Issue #233, opened 2023-10-16
[Epic] Agent dynamic attach


andrewazores:
Currently, the only way to get the Cryostat Agent attached and running on a target application is with the `-javaagent` JVM flag. This means the end user must be able to modify their application launch setup to add this flag, and the user must get the Agent JAR into the application's filesystem somehow, ex. building it into the application image or supplying it in a mounted volume. The user must also decide before launching their application whether they would like to use the Cryostat Agent or not, and if they decide to add the Cryostat Agent later, their application must be restarted.

It should be possible to decouple the Cryostat Agent from these assumptions for the most part.

1. the user might obtain their application image from another team or a separate vendor and may not be able to modify the launcher setup to add a `-javaagent` flag, but would be able to spawn a process within the container or launch a sidecar container
2. the user might be able to supply the Agent as a mounted volume, but not built in to the application image (similar to above). Part of the story here could be an integration with the `cryostat-operator` that can bundle the Agent JAR and supply it as a mounted volume if the user simply adds some annotation to their application Deployment, for example.
3. the user might have a long-running "pet" (not cattle) application and wish to install the Cryostat Agent without taking the application down

If the Agent implements dynamic attach and provides some additional mechanisms to supply its required configurations then the major potential deployment blockers in the above scenarios can be alleviated.

andrewazores:
As seen done by another project recently, the Agent JAR could be built into a simple JAR-carrier container image, ex:

```dockerfile
FROM scratch
cp ./target/cryostat-agent.jar /cryostat
```

This would be an easy way for application images to consume the Agent JAR - rather than manually downloading the JAR to the build machine and copying it into the container build, they could use a multi-stage container build like:

```dockerfile
FROM cryostat/agent-carrier:latest as cryostat-agent
RUN echo '' # no-op, can this be removed?

FROM base-image
COPY --from=cryostat-agent /cryostat/cryostat-agent.jar /app/deployment/lib
... copy application artifacts in
```

----

Alternatively, this JAR-carrier image can be used as an init container in the application's Deployment to achieve the same result dynamically, if there is a volume shared between the init container and the application container:

```
        initContainers:
        - name: download-agent
          image: agent-carrier:latest
          imagePullPolicy: Always
          command: [ 'cp', '/cryostat/cryostat-agent.jar','/etc/agent/ ]
          volumeMounts: # this must also be defined on the container that runs the host JVM which the Agent should attach to, so that the JAR is available on its filesystem as well
[...]

PR #390, opened 2024-04-26
build(image): add Dockerfile for agent carrier container image

andrewazores:
Related to https://github.com/cryostatio/cryostat-agent/issues/233
See https://github.com/cryostatio/test-applications/pull/6 (replaces Quarkus test build from this repo)

```
$ podman build . -f src/main/container/Containerfile -t quay.io/cryostat/cryostat-agent-carrier:latest
```


src/main/container/Containerfile
```diff
@@ -0,0 +1,5 @@
+ARG BASEIMAGE=docker.io/library/busybox:latest
```
ebaron, line 1: What about UBI micro for this? https://catalog.redhat.com/software/containers/ubi8-micro/601a84aadd19c7786c47c8ea

ebaron (commented):
Looks good overall. What about using the name `cryostat-agent-init` instead of carrier to make it clear that this is to be used as an init container only?

ebaron (approved):

PR #397, opened 2024-05-06
feat(CI): Build container image

mwangggg:
related to: https://github.com/cryostatio/cryostat-operator/issues/791

This CI workflow builds the agent image, then packages it in the quarkus-test sample application and pushes it to the crysotat quay repo. 

example workflow in my forked -agent repo: https://github.com/mwangggg/cryostat-agent/actions/runs/8974812712

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -76,7 +76,7 @@ jobs:
     outputs:
       image: ${{ steps.push-to-quay.outputs.image }}
     env:
-      IMAGE_VERSION: ${{ needs.get-pom-properties.outputs.image-version }}
+      IMAGE_VERSION: ${{ needs.build-agent.outputs.image-version }}
     name: Build Java ${{ matrix.java }}
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -96,8 +96,8 @@ jobs:
       uses: redhat-actions/buildah-build@v2
       with:
         image: ${{ env.CI_IMG }}
-        archs: amd64
-        tags: ${{ env.agent-version }} ${{env.DATE_TAG}}
+        archs: amd64, arm64
+        tags: ${{ env.agent-version }} ${{env.DATE_TAG}} latest
         context: quarkus-agent
         containerfiles: |
           ./quarkus-agent/src/main/docker/Dockerfile.jvm
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -90,7 +90,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-build-
           ${{ runner.os }}-
-    - uses: actions/setup-java@v3
+    - uses: actions/setup-java@v4
       with:
         java-version: ${{ matrix.java }}
         distribution: 'temurin'
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -5,38 +5,27 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  #push:
-  #  branches:
-  #    - main
-  #    - v[0-9]+
-  #    - v[0-9]+.[0-9]+
-  #    - cryostat-v[0-9]+.[0-9]+
-  issues:
-    types:
-      - opened
-      - edited
-      - reopened
-  pull_request_target:
-    types:
-      - opened
-      - edited
-      - reopened
-      - labeled
-      - synchronize
+  push:
+    branches:
+      - main
+      - v[0-9]+
+      - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
 
 env:
   OPENSUSE_UNOFFICIAL_LIBCONTAINERS_KEY_URL: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable/xUbuntu_22.04/Release.key"
   OPENSUSE_UNOFFICIAL_LIBCONTAINERS_SOURCE_URL: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable/xUbuntu_22.04"
-  CI_USER: miwan
-  CI_REGISTRY: quay.io/miwan
-  CI_IMG: quay.io/miwan/quarkus-test
+  CI_USER: cryostat+bot
+  CI_REGISTRY: quay.io/cryostat
+  CI_IMG: quay.io/cryostat/quarkus-test
 
 jobs:
   check-before-build:
     runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'cryostatio' }}
     steps:
     - name: Fail if safe-to-test is not applied
-      if: github.repository_owner == 'cryostatio' && (!contains(github.event.pull_request.labels.*.name, 'safe-to-test'))
+      if: (!contains(github.event.pull_request.labels.*.name, 'safe-to-test'))
       run: exit 1
 
   build-agent:
@@ -97,7 +86,7 @@ jobs:
         cache: 'maven'
     - uses: actions/checkout@v4
       with:
-        repository: mwangggg/quarkus-test
+        repository: ${{ github.repository_owner }}/quarkus-test
         ref: main
         submodules: true
         fetch-depth: 0
@@ -148,7 +137,7 @@ jobs:
         tags: ${{ steps.buildah-build.outputs.tags }}
         registry: ${{ env.CI_REGISTRY }}
         username: ${{ env.CI_USER }}
-        password: ${{ secrets.QUAY_PASSWORD }}
+        password: ${{ secrets.REPOSITORY_PASSWORD }}
     - name: store quarkus-test image as output
       id: quarkus-test-image
       run: echo "image=${{ steps.push-to-quay.outputs.registry-path }}" >> $GITHUB_OUTPUT
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -56,7 +56,7 @@ jobs:
       with:
         java-version: '17'
         distribution: 'temurin'
-    - uses: actions/cache@v3
+    - uses: actions/cache@v4
       with:
         path: ~/.m2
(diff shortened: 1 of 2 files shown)


mwangggg:
Screenshot from 2024-05-06 15-06-08
The CI successfully pushed the correct quarkus-test image into my quay repo- I tried it in cryostat3 and it works as expected.
Screenshot from 2024-05-06 15-06-33
The CI also comments the correct image in the PR

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,163 @@
...
+        curl -fsSL $OPENSUSE_UNOFFICIAL_LIBCONTAINERS_KEY_URL | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/devel_kubic_libcontainers_unstable.gpg > /dev/null
+        sudo apt -y purge podman
+        sudo apt update && sudo apt -y install podman
+    - name: Emulate docker with podman
+      run: |
+        mkdir -p $HOME/.bin
+        cat <(echo '#!/usr/bin/env bash') <(echo 'exec podman "$@"') > $HOME/.bin/docker
+        chmod +x $HOME/.bin/docker
+        echo "PATH=$HOME/.bin:$PATH" >> "$GITHUB_ENV"
+    - name: Set up testcontainers for podman
```
andrewazores, line 110: `testcontainers` shouldn't be needed here, that's part of the Quarkus unit/integration test suite but the sample app doesn't use it.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,163 @@
...
+      with:
+        servers: '[{"id": "github", "username": "dummy", "password": "${{ secrets.GITHUB_TOKEN }}"}]'
+        githubServer: false
+    - name: Install podman v4
+      run: |
+        echo "deb $OPENSUSE_UNOFFICIAL_LIBCONTAINERS_SOURCE_URL/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list
+        curl -fsSL $OPENSUSE_UNOFFICIAL_LIBCONTAINERS_KEY_URL | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/devel_kubic_libcontainers_unstable.gpg > /dev/null
+        sudo apt -y purge podman
+        sudo apt update && sudo apt -y install podman
+    - name: Emulate docker with podman
```
andrewazores, line 104: I wonder if this podman setup stuff is needed for this workflow? The `buildah-build` step should hopefully be self-contained enough to be able to perform a build using `buildah` already, which is related to but separate from podman.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,163 @@
...
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+    - uses: actions/setup-java@v4
+      with:
+        java-version: ${{ matrix.java }}
+        distribution: 'temurin'
+        cache: 'maven'
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ github.repository_owner }}/quarkus-test
```
andrewazores, line 89: IMO we shouldn't create a repo like `github.com/cryostatio/quarkus-test`. Something like `github.com/cryostatio/test-applications` would be better, and `quarkus-test` would be a subdirectory there.

Likewise I would also prefer not to push sample application images into `quay.io/cryostat` - maybe https://quay.io/organization/redhat-java-monitoring would be a better fit?

mwangggg: yes these make sense so users don't think the quarkus-test application is actually part of Cryostat. I'm assuming the quay repo would need a different user & password?

andrewazores: Yes the credentials for pushing to that org/repo would be different.

mwangggg:
updated workflow: https://github.com/mwangggg/cryostat-agent/actions/runs/8974812712

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,142 @@
...
+      contents: read
+      pull-requests: write
+      statuses: write
+    outputs:
+      quay-image: ${{ steps.quarkus-test-image.outputs.image }}
+    env:
+      agent-version: ${{ needs.build-agent.outputs.image-version }}
+    name: Build Java ${{ matrix.java }}
+    steps:
+    - uses: actions/checkout@v4
```
andrewazores, line 71: Can this first checkout be removed? I think this would check out the Agent repo, but this job is just going to build the test application and pull the Agent from the last job's local `~/.m2`.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,142 @@
...
+      with:
+        repository: ${{ github.repository_owner }}/test-applications
+        ref: main
+        submodules: true
+        fetch-depth: 0
+    - name: maven-settings
+      uses: s4u/maven-settings-action@v2
+      with:
+        servers: '[{"id": "github", "username": "dummy", "password": "${{ secrets.GITHUB_TOKEN }}"}]'
```
andrewazores, line 96: This looks like it can also be removed - I think this was from using GitHub Maven Packages?

.github/workflows/container-image.yml
mwangggg: It needs to stay there so the right cache is hit- if the pom.xml file is updated and we don't check out the -agent repo, the wrong cache will be hit and an outdated image will be built

andrewazores: Ah, that makes sense.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,137 @@
...
+        image: ${{ env.CI_IMG }}
+        archs: amd64
+        tags: ${{ env.agent-version }}
+        containerfiles: |
+          ./quarkus-test/src/main/docker/Dockerfile.jvm
+    - name: Push to quay.io
+      id: push-to-quay
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        image: quarkus-test
```
andrewazores, line 110: I think this would also be `quarkus-cryostat-agent`, ie the last segment of the `CI_IMG`

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,137 @@
+name: Build Container Image
+
+concurrency:
+  group: ci-${{ github.run_id }}
+  cancel-in-progress: true
+
+on:
```
andrewazores, line 7: Could you add a `workflow_dispatch` event trigger here so we can manually trigger rebuilds? We'll probably want to do that on occasion since the container build just pulls from the `:latest` base image.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,137 @@
...
+        archs: amd64
+        tags: ${{ env.agent-version }}
+        containerfiles: |
+          ./quarkus-agent/src/main/docker/Dockerfile.jvm
+    - name: Push to quay.io
+      id: push-to-quay
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        image: quarkus-cryostat-agent
+        tags: ${{ steps.buildah-build.outputs.tags }}
```
andrewazores, line 111: A `DATE_TAG` would be nice too, especially combined with the `workflow_dispatch`:

https://github.com/cryostatio/cryostat-db/blob/dde25a293cb2e2b2b39e2c7a0f11727387ef0559/.github/workflows/ci.yml#L56

andrewazores:
Looking good. Could you get this to run in your fork to validate that we haven't missed anything among the recent changes? Obviously the quay.io piece won't work, unless you also set up your fork with the token secret.

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,140 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build container images and manifest
+      id: buildah-build
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.CI_IMG }}
+        archs: amd64
+        tags: ${{ env.agent-version }} ${{env.DATE_TAG}}
+        containerfiles: |
+          ./quarkus-agent/src/main/docker/Dockerfile.jvm
```
andrewazores, line 108: https://github.com/redhat-actions/buildah-build?tab=readme-ov-file#inputs-for-build-from-containerfile

I think the build context needs to be supplied here now as well, rather than the default `.` which would be `test-applications`, the context should be `./quarkus-agent`

mwangggg:
I finally got a workflow that works- I completely forgot that there's no easy way to get the PR associated with a push event, so I had to remove the the safe-to-test label checking and commenting the quay image after it's been successfully pushed: https://github.com/mwangggg/cryostat-agent/actions/runs/9813147572

andrewazores:
Ahh, yes, that makes sense. I think it's fine to just have it work on push, I don't think there will be that many PRs going on in this repo.

https://quay.io/repository/redhat-java-monitoring/quarkus-cryostat-agent?tab=tags

image

Workflow results look good. I think it makes sense to tag this container image with the Agent version (as your workflow did) rather than the application's own version, too.

.github/workflows/container-image.yml
andrewazores: Should the DATE_TAG be here too/instead of on the build step above?

mwangggg: The tags in `${{ steps.buildah-build.outputs.tags }}` should include all the tags from the build step above

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,114 @@
...
+      working-directory: quarkus-agent
+      run: ./mvnw -B -U -Dio.cryostat.agent.version=${{ env.agent-version }} clean verify
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build container images and manifest
+      id: buildah-build
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.CI_IMG }}
+        archs: amd64
```
andrewazores, line 99: https://github.com/cryostatio/cryostat-db/blob/dde25a293cb2e2b2b39e2c7a0f11727387ef0559/.github/workflows/ci.yml#L55

Multiarch would be nice for this as well

andrewazores (approved):

Issue #460, opened 2024-08-08
[Task] Add option for users to import whole truststore instead of certs


mwangggg:
^ maybe the Agent should have some other config property to allow it to import a whole truststore as well? Like `cryostat.agent.webclient.tls.truststore.path=/path/to/init/truststore`, `cryostat.agent.webclient.tls.truststore.pass=changeit`. I think this could be used in conjunction with the new configs from https://github.com/cryostatio/cryostat-agent/pull/448  - if these properties aren't specified then it does the same thing as https://github.com/cryostatio/cryostat-agent/pull/448, but if they are specified then probably just [this bit of code](https://github.com/cryostatio/cryostat-agent/blob/938e8f1032415606f5a65cfae20fb377d1cd0b89/src/main/java/io/cryostat/agent/MainModule.java#L180) needs to be adjusted to start with the provided truststore?

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-operator/issues/928#issuecomment-2276490338_
            
Related to https://github.com/cryostatio/cryostat-agent/issues/139

andrewazores:
Sort of on this topic, I tried to use the new cert list config in the environment variable form on OpenShift, and the Deployment was rejected by Kubernetes' environment variable validation because of the `[0]` characters. Maybe we need to come up with some other syntax that can be transformed into an acceptable character set.

mwangggg:
you can add it in as a a jvm system property with JAVA_OPTS_APPEND I think- does that work? We can just tweak the README to inform readers of this change

andrewazores:
That's what I had to do in the Operator PR. We can't assume that the `JAVA_OPTS_APPEND` mechanism is available for all end users' applications, though there is likely to be at least some equivalent. It would still be nicest to have a format for this property so that it can be specified as either environment variable or system property, since all the others can be specified either way as well.

andrewazores:
https://stackoverflow.com/questions/68862008/how-to-convert-a-collection-in-config-to-environment-variable-in-microprofile-qu

PR #461, opened 2024-08-09
feat(TLS): allow users to import whole truststore + pass

mwangggg:
fixes: #460


src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -176,11 +181,19 @@ public X509Certificate[] getAcceptedIssuers() {
...
+                    ts.load(truststore, truststorePass.get().toCharArray());
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+
+            // initialize truststore with user provided certs
+            for (TruststoreConfig truststore : truststoreCerts) {
                 // load truststore with certificatesCertificate
                 InputStream certFile = new FileInputStream(truststore.getPath());
```
andrewazores, line 198: I missed this the first time around - I think this could be done with autoclose, ie `try (InputStream certFile = ...)`, to be sure it's properly released.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -176,11 +181,19 @@ public X509Certificate[] getAcceptedIssuers() {
                 }
             }
 
-            // initialize truststore
-            KeyStore ts = KeyStore.getInstance(KeyStore.getDefaultType());
-            ts.load(null, null);
```
andrewazores, line 181: Should this still `.load(null, null)` if no truststore config is provided?

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -251,12 +266,75 @@ public static int provideCryostatAgentWebclientResponseTimeoutMs(Config config)
...
+        try (FileInputStream passFile = new FileInputStream(truststorePassFile.get())) {
+            String pass = IOUtils.toString(passFile, Charset.forName(passCharset));
+            pass = pass.substring(0, pass.length() - 1);
+            password = Optional.ofNullable(pass);
+        } catch (NoSuchElementException e) {
+            return password;
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return password;
```
andrewazores, line 295: We should be careful with the passwords here. I think this is going to result in the password (or passfile content) getting left with a dangling reference in the dependency injection system, which will leave the password contents allocated on the heap forever.

It would be better if this was something like `Optional<ByteBuffer>`, where

```java
public class ByteBuffer {
  private final byte[] buf;
  public ByteBuffer(int len) {
    this.buf = new byte[len];
  }
  public void set(String s) {
    assert s.length() <= this.buf.length;
    Arrays.copy(s.getBytes(), this.buf);
  }
  public String get() {
    return new String(this.buf);
  }
  public void clear() {
    Arrays.fill(this.buf, '\0');
  }
```

(rough pseudojava)

This way `ByteBuffer#clear()` can be called after we're done initializing the truststore and the password is only present on the heap for the shortest amount of time possible. It will still be present somewhere in the environment, as an environment variable or system property etc., but we can minimize the surface area.

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -254,12 +268,77 @@ public static int provideCryostatAgentWebclientResponseTimeoutMs(Config config)
...
+            @Named(CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS_CHARSET) String passCharset) {
+        Optional<String> truststorePassFile =
+                config.getOptionalValue(
+                        CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS_FILE, String.class);
+        if (truststorePassFile.isEmpty()) {
+            return Optional.empty();
+        }
+        try (FileInputStream passFile = new FileInputStream(truststorePassFile.get())) {
+            String pass = IOUtils.toString(passFile, Charset.forName(passCharset)).trim();
+            return Optional.ofNullable(new ByteBuffer(pass, passCharset));
```
andrewazores, line 292: Is there some way to go directly from `FileInputStream` into `ByteBuffer`, without reading out the file contents with ex. `IOUtils.toString()`? This is still somewhat vulnerable to the problem I outlined before - this `String pass` gets allocated on the heap and contains the password, and when this variable drops out of scope it becomes eligible for garbage collection but will remain on the heap until that happens, which can take some time (or never if the host JVM happens to be using EpsilonGC). At least now it isn't held on the heap indefinitely, but it seems there could still be room to improve this by reading bytes directly from the input stream and copying them into the buffer's `byte[]`, so that they can be explicitly and deterministically cleared.

mwangggg: ohh hmm I will take a look

andrewazores: https://stackoverflow.com/questions/1264709/convert-inputstream-to-byte-array-in-java

I think a simple `passFile.readAllBytes()` or `passFile.toByteArray()` would do the trick here, plus a little refactoring of the `ByteBuffer`.

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -308,13 +306,14 @@ public static String provideCryostatAgentWebclientTlsTruststorePassCharset(Confi
     public static Optional<ByteBuffer> provideCryostatAgentWebclientTlsTruststorePass(
             Config config,
             @Named(CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS_FILE)
-                    Optional<ByteBuffer> truststorePass) {
+                    Optional<ByteBuffer> truststorePass,
+            @Named(CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS_CHARSET) String passCharset) {
         Optional<String> opt =
                 config.getOptionalValue(CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS, String.class);
```
andrewazores, line 312: I don't suppose swapping `String.class` for `byte[].class` is supported by microprofile-config, is it?

mwangggg: I can make a converter so in ConfigModule I can extract `byte[]` or even `ByteBuffer` type, but it would still use `config.getOptionalValue(CRYOSTAT_AGENT_WEBCLIENT_TLS_TRUSTSTORE_PASS, String.class` in the converter itself.

andrewazores: I figured. Not worth doing in that case - I'm just trying to see if there's any feasible way for us to get around having to ever read this value as a String.

Anyway, supplying the password as a mounted volume/projected file is already the better security practice, so if we're handling that case properly with a cleared byte array, then I think the bases are covered for the real security aspects.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -177,24 +185,41 @@ public X509Certificate[] getAcceptedIssuers() {
...
-            KeyStore ts = KeyStore.getInstance(KeyStore.getDefaultType());
+            KeyStore ts = KeyStore.getInstance(truststoreType);
             ts.load(null, null);
 
-            for (TruststoreConfig truststore : truststores) {
+            // initialize truststore with user provided path and pass
+            if (!truststorePath.isEmpty() && !truststorePass.isEmpty()) {
+                try (InputStream truststore = new FileInputStream(truststorePath.get())) {
+                    ts.load(truststore, truststorePass.get().get(passCharset).toCharArray());
+                    truststorePass.get().clear();
```
andrewazores, line 195: Move this `.clear()` call into a `finally` clause, please, to ensure it's blanked out even if the truststore fails to load.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -177,24 +185,41 @@ public X509Certificate[] getAcceptedIssuers() {
...
-            // initialize truststore
-            KeyStore ts = KeyStore.getInstance(KeyStore.getDefaultType());
+            KeyStore ts = KeyStore.getInstance(truststoreType);
             ts.load(null, null);
 
-            for (TruststoreConfig truststore : truststores) {
+            // initialize truststore with user provided path and pass
+            if (!truststorePath.isEmpty() && !truststorePass.isEmpty()) {
+                try (InputStream truststore = new FileInputStream(truststorePath.get())) {
+                    ts.load(truststore, truststorePass.get().get(passCharset).toCharArray());
```
andrewazores, line 194: `.get().get(charset).toCharArray()` passes the bytes into a `new String()`, which would also be great to avoid doing if possible. Is there some other way to do the byte->char mapping using the charset without passing through a String constructor?

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -642,4 +719,28 @@ public static URIRange fromString(String s) {
...
+
+    public static class ByteBuffer {
+        private final byte[] buf;
+
+        public ByteBuffer(int len) {
+            this.buf = new byte[len];
+        }
+
+        public ByteBuffer(byte[] s) {
+            this.buf = Arrays.copyOf(s, s.length);
```
andrewazores, line 731: Should be careful here - if we're just copying the array in, then we are able to control the new copy and clear it as needed, but the original one is still potentially floating around out there containing the contents indefinitely. Wherever this constructor is invoked should take care to clear the original array too once the ByteBuffer has been constructed.

andrewazores (approved):

PR #492, opened 2024-09-24
feat(diagnostics): Add API handler for invoking GC

Josh-Matsuoka:
Hi,

Related to https://github.com/cryostatio/cryostat/issues/133

Adds an API handler to invoke GC on a target through the Memory MBean. 

Opening as Draft while I adjust it to the approach/endpoint expected by https://github.com/cryostatio/cryostat/pull/644/


src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /*
+    * public <T> T invokeMBeanOperation(
+           String beanName,
+           String operation,
+           Object[] parameters,
+           String[] signature,
+           Class<T> returnType)
+    */
```
andrewazores, line 53: Remove?

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    static class MBeanInvocationRequest<T> {
+
+        public String beanName;
+        public String operation;
+        public Object[] parameters;
+        public String[] signature;
+        public Class<T> returnType;
```
andrewazores, line 91: I don't think the `returnType` is needed on this side - it's used on the Cryostat server side so that the internal callers can indicate how the response body should be parsed, but here on the Agent side I think that if the M(X)Bean operation being invoked has any return value, then we should just serialize that out as JSON or whatever format is applicable. If we want there to be some way for the server to request a particular format then that should primarily be done using HTTP content negotiation, ie the `Accept` request header.

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,110 @@
...
+
+        // Support GC and Thread operations for now
+        public boolean isValid() {
+            if (this.beanName.equals(ManagementFactory.MEMORY_MXBEAN_NAME)
+                    || this.beanName.equals(ManagementFactory.THREAD_MXBEAN_NAME)) {
```
andrewazores, line 96: Let's remove the Thread bean for now, since I wasn't able to get anything useful working with that yet.

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+        try {
+            String mtd = exchange.getRequestMethod();
+            switch (mtd) {
+                case "POST":
+                    try (InputStream body = exchange.getRequestBody()) {
+                        MBeanInvocationRequest req =
+                                mapper.readValue(body, MBeanInvocationRequest.class);
+                        if (req.beanName.equals(ManagementFactory.MEMORY_MXBEAN_NAME)) {
```
andrewazores, line 64: Rather than having an if-else, or a `switch`, or other mechanism for enumerating beans, can't we do an MXBean lookup by name using the `req.beanName` and then pass the other `req` fields to `invoke`?

```java
var mbeanServer = ManagementFactory.getPlatformMBeanServer();
if (!req.isValid()) // respond HTTP 400 Bad Request
Object response = mbeanServer
    .invoke(ObjectName.getInstance(req.beanName), req.operation, req.parameters, req.signature);
// respond HTTP 200 with response body JSON, or HTTP 201 if response is null
```

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,110 @@
...
+
+import javax.inject.Inject;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sun.net.httpserver.HttpExchange;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class InvokeContext implements RemoteContext {
```
andrewazores, line 31: I think this should be a `MutatingRemoteContext`, so that any M(X)Bean operations can only be invoked if the agent has been configured to allow "API writes". That is a bit of a misnomer now since a GC operation isn't exactly a write operation, but this remote invocation context in general opens the possibility of mutation operations

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,118 @@
...
+                                        req.operation,
+                                        req.parameters,
+                                        req.signature);
+                        if (Objects.nonNull(response)) {
+                            exchange.sendResponseHeaders(HttpStatus.SC_OK, BODY_LENGTH_UNKNOWN);
+                            try (OutputStream responseStream = exchange.getResponseBody()) {
+                                mapper.writeValue(responseStream, response);
+                            }
+                        } else {
+                            exchange.sendResponseHeaders(HttpStatus.SC_CREATED, BODY_LENGTH_NONE);
```
andrewazores, line 77:
```suggestion
                            exchange.sendResponseHeaders(HttpStatus.SC_ACCEPTED, BODY_LENGTH_NONE);
```

Thinking about it a bit more, I think 202 might be a better code for this condition than 201. Either way, it's a success code that is distinguishable from 200, so I don't think it's critically important. We might even want to return 201/202/204 depending on the exact operation being invoked later on, for example.

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    static class MBeanInvocationRequest<T> {
+
+        public String beanName;
+        public String operation;
+        public Object[] parameters;
+        public String[] signature;
+
+        // Support GC and Thread operations for now
```
andrewazores, line 102:
```suggestion
```

andrewazores (commented):
- https://github.com/cryostatio/cryostat-agent/pull/492
- https://github.com/cryostatio/cryostat/pull/644
- https://github.com/cryostatio/cryostat-core/pull/459
- https://github.com/cryostatio/cryostat-web/pull/1426

all seem to be working well when tested together. The UI needs a bit of tweaking so I'll hold off approving/merging the backend pieces for now, but I think they are all OK.

andrewazores (approved):

PR #491, opened 2024-09-23
feat(tls): agent can pass TLS client auth cert

andrewazores:
Fixes https://github.com/cryostatio/cryostat-agent/issues/490
See also https://github.com/cryostatio/cryostat-operator/pull/957

See above PR for steps on retrieving `tls.crt`, `tls.key`, `ca.crt` from the Operator's generated Secret for the Agent to use.

Then, to test this combined with the above Operator PR:

1. Patch the `sample-app.yaml` to deploy the container `quay.io/andrewazores/quarkus-cryostat-agent:tls-client-auth-15` (an application containing the latest build of this PR)
2. `make sample_app`
3. `export SAMPLE_POD="$(kubectl get pod -l app=quarkus-test -o jsonpath='{$.items[0].metadata.name}')" ; oc wait pod $SAMPLE_POD --for condition=ready; for i in tls.key tls.crt ca.crt; do oc cp $i $SAMPLE_POD:/tmp/ ; done` to get the secret contents into the Agent application's Pod
4. In another terminal `oc wait pod -l app=quarkus-test --for condition=ready ; oc logs -f -l app=quarkus-test` to watch the log output from the sample application Pod. Initially the Agent should start up but bail out because the `baseuri` property is empty. This is expected.
[...]

PR #507, opened 2024-10-04
feat(tls): add configuration options for supplying bare TLS server key

andrewazores:
Based on https://github.com/cryostatio/cryostat-agent/pull/491
See https://github.com/cryostatio/cryostat-agent/pull/491#pullrequestreview-2346679619

> On a side note, would it be possible to add a key options for the web server in addition to the existing keystore options, similar to what you've done here for client-auth? It would make it easier to configure HTTPS for the server.


andrewazores:
I tried to mimic the "bare key file" approach that I did for client authentication, but hook it in to the existing webserver HTTPS setup. The keystore file can be provided alone, or the bare key can be provided alone in which case a new empty keystore is generated to contain it. If both are specified then the keystore file will be loaded, then the additional bare key will be inserted into it.

ebaron (approved):

PR #508, opened 2024-10-04
feat(versions): record own build info/version as properties, validate server version

### LINKED ISSUE #504: [Request] Cryostat server version validation
When the Agent starts up, it should fire a `GET /health` request to the Cryostat server and parse out the server's version. The Agent should have build-time properties to set the minimum and maximum server versions that it expects to work with, and if the server version falls outside this range, the Agent should print a warning.

### LINKED ISSUE #505: [Request] Build info properties
See #504

The Agent build should embed some build information into the JAR, such as the pom.xml version and git commit. This data should be written into a .properties file which gets bundled into the JAR. The Agent can load this file from the class path at startup and ensure these are set as System properties. These will then be visible through the MBean Metrics interface, for example on the Cryostat Web UI's JVM Details component.

This mechanism can also be how #504 is implemented - the server version range can be set in the pom.xml, and loaded at runtime this way

andrewazores:
Fixes #504
Fixes #505

At build time, the following bits of information are recorded into resources files which are included in the distribution JAR:
- `pom.xml` `version`, ie the Agent's own version string
- `pom.xml` `cryostat.server.version.min` and `cryostat.server.version.max`, the minimum (inclusive) and maximum (exclusive) Cryostat server versions that this Agent version is expected to work with. Currently set to 4.0.0 and 5.0.0 respectively, forming the range `[4.0.0, 5.0.0)`, so that any Cryostat server version `4.y.z` should be within range. These are set as `pom.xml` properties rather than in the `microprofile-config.properties` so that they are build-time immutable.
- current git commit hash

The Agent reads its own version string and logs this at startup. It loads all of these values and sets them as various system properties so that they can be retrieved generically later on. Finally, when it attempts to register itself with the Cryostat server, it performs a `GET /health` check against the server to obtain the server's `cryostatVersion` string, then parses that to a semantic version, and compares that to the expected server version range. If the actual server version is outside of the expected range then the agent logs a warning.

```
2024-10-04 18:07:44:382 +0000 [cryostat-agent-main] INFO io.cryostat.agent.Agent - Cryostat Agent version 0.5.0-SNAPSHOT starting...
...
2024-10-04 18:08:25:018 +0000 [cryostat-agent-worker-2] DEBUG io.cryostat.agent.Registration - Connected to Cryostat server: version 4.0.0 , build abfc597f1a7ad93093f7ead616766d56b4333809
```

Screenshot_2024-10-04_14-12-10

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,8 +104,12 @@
   <maven-source-plugin.version>3.3.1</maven-source-plugin.version>
   <maven-javadoc-plugin.version>3.10.1</maven-javadoc-plugin.version>
   <jreleaser-maven-plugin.version>1.14.0</jreleaser-maven-plugin.version>
+  <org.codehaus.mojo.exec.plugin.version>3.4.1</org.codehaus.mojo.exec.plugin.version>
 
   <shade.prefix>io.cryostat.agent.shaded</shade.prefix>
+
+  <cryostat.server.version.min>4.0.0</cryostat.server.version.min>
+  <cryostat.server.version.max>5.0.0</cryostat.server.version.max>
 </properties>
 
 <dependencyManagement>
@@ -298,6 +302,29 @@
           </manifestEntries>
         </archive>
       </configuration>
+    </plugin>
+    <plugin>
+      <groupId>org.codehaus.mojo</groupId>
+      <artifactId>exec-maven-plugin</artifactId>
+      <version>${org.codehaus.mojo.exec.plugin.version}</version>
+      <executions>
+        <execution>
+          <id>generate-git-version</id>
+          <phase>generate-resources</phase>
+          <goals>
+            <goal>exec</goal>
+          </goals>
+          <configuration>
+            <executable>git</executable>
+            <arguments>
+              <argument>rev-parse</argument>
+              <argument>--verify</argument>
+              <argument>HEAD</argument>
+            </arguments>
+            <outputFile>${project.build.directory}/classes/META-INF/gitinfo</outputFile>
+          </configuration>
+        </execution>
+      </executions>
     </plugin>
     <plugin>
       <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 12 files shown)


Josh-Matsuoka (approved):
I think the fixme in BuildInfo is probably worthwhile, we should open an issue for core to keep track of it. Other than that LGTM

ebaron:
I'm getting an error upon agent startup after this change:
```
Exception in thread "cryostat-agent-main" Exception in thread "cryostat-agent-main" java.lang.NumberFormatException: For input string: "0-SNAPSHOT"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:668)
	at java.base/java.lang.Integer.parseInt(Integer.java:786)
	at io.cryostat.agent.VersionInfo$Semver.fromString(VersionInfo.java:109)
	at io.cryostat.agent.VersionInfo.load(VersionInfo.java:50)
	at io.cryostat.agent.Agent.accept(Agent.java:210)
	at io.cryostat.agent.Agent.lambda$agentmain$0(Agent.java:160)
	at java.base/java.lang.Thread.run(Thread.java:840)
java.lang.NumberFormatException: For input string: "0-SNAPSHOT"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:668)
	at java.base/java.lang.Integer.parseInt(Integer.java:786)
	at io.cryostat.agent.VersionInfo$Semver.fromString(VersionInfo.java:109)
	at io.cryostat.agent.VersionInfo.load(VersionInfo.java:50)
	at io.cryostat.agent.Agent.accept(Agent.java:210)
	at io.cryostat.agent.Agent.lambda$agentmain$0(Agent.java:160)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

PR #524, opened 2024-10-28
feat(config): add config properties for callback URL components

ebaron:
This work is done to support the [agent auto-configuration](https://github.com/cryostatio/cryostat-operator/issues/784) work in the operator. Specifically, this is needed to support generating wildcard certificates for the agent to use.

To test:
1. Deploy Cryostat using the operator.
2. Deploy the sample app w/ agent.
3. Swap out the default container image with one built using this PR.
4. Create a headless service:
    ```yaml
    apiVersion: v1
    kind: Service
    metadata:
      name: headless
      namespace: cryostat-operator-system
    spec:
      clusterIP: None
      selector:
        app: quarkus-cryostat-agent
      type: ClusterIP
    ```
5. Add a `POD_NAME` environment variable
    ```
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    ```
5.  Add these system properties to `JAVA_OPTS_APPEND`:
    ```
    -Dcryostat.agent.callback.scheme=http
    -Dcryostat.agent.callback.host-name=$(POD_NAME),$(POD_IP)[replace("."\,"-")]
    -Dcryostat.agent.callback.domain-name=headless.cryostat-operator-system.svc
    -Dcryostat.agent.callback.port=9977
    -Dio.cryostat.agent.shaded.org.slf4j.simpleLogger.defaultLogLevel=debug
    ```
6. Observe a lines similar to:
    ```
    2024-11-04 18:41:56:456 +0000 [cryostat-agent-main] DEBUG io.cryostat.agent.ConfigModule - Attempting to resolve quarkus-cryostat-agent-5cd54bbb86-62zsv.headless.cryostat-operator-system.svc
    2024-11-04 18:41:56:546 +0000 [cryostat-agent-main] DEBUG io.cryostat.agent.ConfigModule - Attempting to resolve 10-128-0-67.headless.cryostat-operator-system.svc
    2024-11-04 18:41:56:550 +0000 [cryostat-agent-main] DEBUG io.cryostat.agent.ConfigModule - Resolved 10-128-0-67.headless.cryostat-operator-system.svc to 10.128.0.67
    2024-11-04 18:41:56:552 +0000 [cryostat-agent-main] DEBUG io.cryostat.agent.ConfigModule - Using http://10-128-0-67.headless.cryostat-operator-system.svc:9977 for callback URL
    ```

Fixes: #525


src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -968,4 +985,65 @@ public void clear() {
...
+            }
+
+            try {
+                URI result =
+                        new URIBuilder()
+                                .setScheme(k8sScheme.get())
+                                .setHost(resolvedHost.get())
+                                .setPort(k8sPort.get())
+                                .build();
+                log.debug("Using " + result.toASCIIString() + " for callback URL");
```
andrewazores, line 1031:
```suggestion
                log.debug("Using {} for callback URL", result.toASCIIString());
```

src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -968,4 +985,65 @@ public void clear() {
...
+                throw new IllegalArgumentException(e);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static boolean tryResolveHostname(String hostname) {
+        try {
+            InetAddress addr = InetAddress.getByName(hostname);
+            log.debug("Resolved " + hostname + " to " + addr.getHostAddress());
```
andrewazores, line 1043:
```suggestion
            log.debug("Resolved {} to {}", hostname, addr.getHostAddress());
```

README.md
```diff
@@ -249,6 +249,11 @@ and how it advertises itself to a Cryostat server instance. Properties that requ
...
+- [ ] `cryostat.agent.kubernetes.callback.scheme` [`String`]: A Kubernetes-specific override for the scheme portion of the `cryostat.agent.callback` URL (e.g. `https`).
+- [ ] `cryostat.agent.kubernetes.callback.pod.name` [`String`]: A Kubernetes-specific override for the host portion of the `cryostat.agent.callback` URL. If this pod is resolvable using its name as a host name, that will be used in the callback URL.
+- [ ] `cryostat.agent.kubernetes.callback.ip` [`String`]: A Kubernetes-specific override for the host portion of the `cryostat.agent.callback` URL. If this pod is resolvable using the dashed IPv4 address (e.g. 1-2-3-4) as a host name, that will be used in the callback URL.
+- [ ] `cryostat.agent.kubernetes.callback.domain` [`String`]: A Kubernetes-specific override for the domain portion of the `cryostat.agent.callback` URL. This will be appended to a resolvable host name to form the callback URL.
+- [ ] `cryostat.agent.kubernetes.callback.port` [`int`]: A Kubernetes-specific override for the port portion of the `cryostat.agent.callback` URL.
```
andrewazores, line 256: I wonder if there is some way we could make this a little more generic. I worry that this works well in the k8s CoreDNS case, but might run into similar extensibility issues as the original callback URL property, on non-CoreDNS systems, or else in systems like Docker Swarm or other things entirely.

It seems like the URI components we need are really:
1. scheme
2. port
3. hostname segments

The scheme and port are easy and we can expect those to be fixed single values. The key piece here is the hostname segment part, where in this particular scenario we have a fixed suffix (the domain portion) and some prefix options (the pod name or IP). This part I think can be made generic quite easily as a simple list-of-lists: `[[DASHED_POD_IP, POD_NAME], [DOMAIN]]` could be processed to test each permutation and use the first resolvable one, the same as what your implementation now does.

The next problem comes in how we get from the IPv4 address to the dashed format. This part seems pretty k8s-specific, or CoreDNS-specific, so 1) this behaviour shouldn't be hardcoded in a generic solution, 2) a generic solution should allow the user to specify what kind of transformation(s) to apply, somehow.

Luckily we're already using CEL, and I think that's a reasonably safe and powerful way to go here:

https://github.com/projectnessie/cel-java/blob/d5db94c58714cffc1d8bfe091da412837b47550e/core/src/main/java/org/projectnessie/cel/extension/StringsLib.java#L136

So I would propose config properties like this:

1. `cryostat.agent.callback [String]` - the original config property for the callback URI. I think if this is defined then it should take precedence over the piecewise configuration that follows.
2. `cryostat.agent.callback.scheme [String]` - ex. `http`
3. `cryostat.agent.callback.port [int]` - ex. `1234`
4. `cryostat.agent.callback.host-segments [List<String>]` - ex. `[[$POD_NAME, $POD_IP], ["headless"], ["cryostat-operator-system"], ["svc"]]`, where `$POD_NAME` and `$POD_IP` are environment variables substituted by the deployment, so these are actually string values passed to the configuration property. The microprofile config library [knows how to automatically convert comma-separated values into lists](https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#_config_value_conversion_rules), so I suggest using a delimiter character that is not URI-legal for creating the sub-lists: `$POD_IP|$POD_NAME,headless,cryostat-operator-system,svc`. At the injection point the resulting `List<String>` can be easily processed into `List<List<String>>` by simply splitting each element on `|`. I think we can also use some optional syntax here for defining CEL expression transformations that can be applied to each segment, for example `$POD_IP[replace(".", "-")]|$POD_NAME, headless, cryostat-operator-system, svc`.

Let me know what you think of this. I can help write it out too, since I realize it's a bit more complex.

ebaron: I think that seems reasonable, but does complicate the implementation a bit. In the above example, would Microprofile split on the comma inside `replace(".", "-")` as well?

andrewazores: Hmm, that's a good question. It might. I'd hope there's some way to escape the comma so that it doesn't do that.

ebaron: I've implemented this mostly using the syntax you suggested, but in order to avoid some n-factor Cartesian product, I retained the separate host-name and domain-name properties. The domain name works as before, while the host name accepts a list, where each item can be transformed by an optional CEL expression. What do you think?

andrewazores: I think this should work well enough. The `host + "." + domain` construction should be OK even when only the hostname is needed for resolution:

```jshell
$ jshell
|  Welcome to JShell -- Version 21.0.5
|  For an introduction type: /help intro

jshell> InetAddress.getByName("localhost.")
$1 ==> localhost./127.0.0.1
```

The docs (README) should be explicit that the CEL transformations are only available for the hostname, and not for the domain name portion. Hopefully this is not too much of a restriction, but I can't really think of any particular case where the domain name would need to be dynamically calculated. In a cluster-internal k8s case it's likely that the domain is just going to be something like `mynamespace.svc.cluster.local` anyway, so simple environment variable substitution on the namespace should get us there.

andrewazores:
Everything else looks good, just the little note about the README above.

README.md
ebaron: Updated the README. How does it look now?

andrewazores (approved):

Issue #509, opened 2024-10-08
[Request] Agent should require TLS connection to server by default


andrewazores:
The Agent should check its `baseuri` property that points to the Cryostat server with which it should register. By default, the Agent should enforce that this URI uses an `https` scheme, and when the Agent opens a connection, it should expect that it really does find an HTTPS connection, with a TLS certificate that it trusts, etc.

It should be possible to disable this and allow insecure HTTP (mainly for testing/prototyping purposes), but this should be controlled by a new and separate configuration property.

PR #554, opened 2024-12-18
feat(tls): Use TLS by default in the agent

Josh-Matsuoka:
Hi,

this PR enforces TLS by default in the agent, controlled by a new configuration option cryostat.agent.tls.enabled

Fixes https://github.com/cryostatio/cryostat-agent/issues/509


README.md
```diff
@@ -255,6 +255,7 @@ and how it advertises itself to a Cryostat server instance. Properties that requ
 - [ ] `cryostat.agent.callback.port` [`int`]: An override for the port portion of the `cryostat.agent.callback` URL.
 - [ ] `rht.insights.java.opt-out` [`boolean`]: for the Red Hat build of Cryostat, set this to true to disable data collection for Red Hat Insights. Defaults to `false`. Red Hat Insights data collection is always disabled for community builds of Cryostat.
 - [ ] `rht.insights.java.debug` [`boolean`]: for the Red Hat build of Cryostat, set this to true to enable debug logging for the Red Hat Insights Java Agent. Defaults to `false`. Red Hat Insights data collection is always disabled for community builds of Cryostat.
+- [ ] `cryostat.agent.tls.enabled` [`boolean`]: Specify wether the agent should use TLS by default, expecting the base URI to be an https connection with a certificate it trusts. Defaults to `true`. Should only be disabled for testing/prototyping purposes.
```
andrewazores, line 258: I think this might make sense to name like `cryostat.agent.webclient.tls.required`, or something like that. It seems like it fits in well with the other existing `cryostat.agent.webclient.tls.*` properties.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -236,10 +236,20 @@ public static SSLContext provideClientSslContext(
...
             @Named(ConfigModule.CRYOSTAT_AGENT_WEBCLIENT_TLS_CLIENT_AUTH_KEY_MANAGER_TYPE)
-                    String clientAuthKeyManagerType) {
+                    String clientAuthKeyManagerType,
+            @Named(ConfigModule.CRYOSTAT_AGENT_BASEURI) URI baseUri,
+            @Named(ConfigModule.CRYOSTAT_AGENT_TLS_ENABLED) boolean tlsEnabled) {
         try {
             KeyManager[] keyManagers = null;
-            if (clientAuthCertPath.isPresent() && clientAuthKeyPath.isPresent()) {
+            if (clientAuthCertPath.isPresent() && clientAuthKeyPath.isPresent() && tlsEnabled) {
+                if (!baseUri.getScheme().contains("https")) {
```
andrewazores, line 245: Should this be `.equals()` rather than `contains()`?

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -674,10 +698,11 @@ public static HttpServer provideHttpServer(
             ScheduledExecutorService executor,
             @Named(ConfigModule.CRYOSTAT_AGENT_WEBSERVER_HOST) String host,
             @Named(ConfigModule.CRYOSTAT_AGENT_WEBSERVER_PORT) int port,
-            @Named(HTTP_SERVER_SSL_CTX) Optional<SSLContext> sslContext) {
+            @Named(HTTP_SERVER_SSL_CTX) Optional<SSLContext> sslContext,
+            @Named(ConfigModule.CRYOSTAT_AGENT_TLS_ENABLED) boolean tlsEnabled) {
         try {
             HttpServer http;
-            if (sslContext.isEmpty()) {
+            if (sslContext.isEmpty() || !tlsEnabled) {
```
andrewazores, line 705: This is for configuring the Agent's own built-in webserver, whereas the issue and the rest of the PR is about the Agent's HTTP client connection to the Cryostat server. I think this part should be removed - these are two different contexts and should be controlled separately.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -236,10 +236,20 @@ public static SSLContext provideClientSslContext(
...
                     String clientAuthKeyPassFileCharset,
             @Named(ConfigModule.CRYOSTAT_AGENT_WEBCLIENT_TLS_CLIENT_AUTH_KEY_MANAGER_TYPE)
-                    String clientAuthKeyManagerType) {
+                    String clientAuthKeyManagerType,
+            @Named(ConfigModule.CRYOSTAT_AGENT_BASEURI) URI baseUri,
+            @Named(ConfigModule.CRYOSTAT_AGENT_WEBCLIENT_TLS_REQUIRED) boolean tlsEnabled) {
         try {
             KeyManager[] keyManagers = null;
-            if (clientAuthCertPath.isPresent() && clientAuthKeyPath.isPresent()) {
+            if (clientAuthCertPath.isPresent() && clientAuthKeyPath.isPresent() && tlsEnabled) {
```
andrewazores, line 244: I think the `tlsEnabled` check should be separated out. The variable/parameter should probably also be renamed `tlsRequired` to match the configuration property name, for clarity of reading.

The `clientAuthCertPath` and `clientAuthKeyPath` are there to enable configuration of the Agent's HTTP Client instance that it uses for talking to the Cryostat server. So I think the check we want is like:

```
if (tlsRequired) {
  if (!baseUri.getScheme().equals("https")) throw new IllegalArgumentException("Base URI must be https");
}
```

This way, if the Agent has TLS enforcement turned on, and the user configures it with a non-HTTPS server base URI, then the Agent will refuse to make that connection. If the user has configured with with an HTTPS server base URI, but did not configure the client auth cert/key paths, then the connection should already fail at a later point because the TLS auth will fail.

src/main/java/io/cryostat/agent/MainModule.java
```diff
@@ -302,13 +312,17 @@ public static SSLContext provideClientSslContext(
...
-            } else if (clientAuthCertPath.isPresent() || clientAuthKeyPath.isPresent()) {
+            } else if (clientAuthCertPath.isPresent()
+                    || clientAuthKeyPath.isPresent()
+                    || tlsEnabled) {
                 throw new IllegalArgumentException(
                         String.format(
                                 "To use TLS client authentication, both the certificate (%s) and"
-                                        + " private key (%s) properties must be set.",
+                                    + " private key (%s) properties must be set. The (%s) property"
+                                    + " must be true as well.",
```
andrewazores, line 322: Here as well, I think the `tlsEnabled` (`tlsRequired`) condition needs to be treated separately from the TLS auth cert/key configuration.

I think we just want to enforce that TLS (HTTPS) is used, not necessarily that *TLS client certificate authentication* is used.

andrewazores (approved):

PR #619, opened 2025-03-31
feat(triggers): Implement smart triggers from configuration files

Josh-Matsuoka:
Addresses https://github.com/cryostatio/cryostat-agent/issues/227

Adds a new configuration property cryostat.agent.smart-trigger.config.path that defines a path for the agent to search for files containing smart trigger definitions.

On agent load when the smart triggers system starts it will scan this directory and parse smart triggers from the files it finds. Mirrors the work done in other areas of Cryostat for declarative configurations.


src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -35,10 +41,50 @@ public class TriggerParser {
...
     }
 
+    public List<SmartTrigger> parseFromFiles() {
+        List<SmartTrigger> triggers = new ArrayList<SmartTrigger>();
+        if (!checkDir()) {
+            log.warn(
+                    "Configuration directory "
+                            + triggerPath.toString()
+                            + " doesn't exist or is missing permissions");
+            return Collections.emptyList();
```
andrewazores, line 58: May as well just return `triggers` here since it will still be empty by this point, unless you also refactor the stream handling below to get rid of `triggers`.

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -35,10 +41,50 @@ public class TriggerParser {
...
+        } catch (IOException e) {
+            log.error(e.getMessage());
+        }
+        return triggers;
+    }
+
+    private List<SmartTrigger> createFromFile(java.nio.file.Path path) {
+        try {
+            String triggerDefinitions = Files.readString(path);
+            return parse(triggerDefinitions);
```
andrewazores, line 74: Since this just passes through to the existing `parse` method it expects that the file content will be just like the previous system property or environment variable value format. I think this is a missed opportunity to allow the definitions provided by files to be a little more readable, by using newline separators rather than commas. Maybe a `java.util.Scanner` or something would be a useful addition here as an additional pass on the raw file contents before handing over to the `parse` method?

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -35,10 +43,52 @@ public class TriggerParser {
...
+        } catch (IOException e) {
+            log.error(e.getMessage());
+            return Collections.emptyList();
+        }
+    }
+
+    private List<SmartTrigger> createFromFile(java.nio.file.Path path) {
+        try {
+            String triggerDefinitions = Files.readString(path);
+            return Arrays.asList(triggerDefinitions.split("\n")).stream()
```
andrewazores, line 76: Probably best to use `System.lineSeparator()` here, just in case.

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -35,10 +43,52 @@ public class TriggerParser {
...
+    java.nio.file.Path triggerPath;
+
     public TriggerParser(FlightRecorderHelper flightRecorderHelper) {
         this.flightRecorderHelper = flightRecorderHelper;
     }
 
+    public List<SmartTrigger> parseFromFiles() {
+        if (!checkDir()) {
+            log.warn(
+                    "Configuration directory {}" + " doesn't exist or is missing permissions",
```
andrewazores, line 56: needless `+` break here now

andrewazores (approved):

PR #635, opened 2025-04-23
fix(callback): perform callback resolution within registration loop

### LINKED ISSUE #634: [Bug] DNS resolution of callback only happens once at startup
Part of the recent autoconfiguration feature is that the Agent can be configured to dynamically determine its own callback URL from components and a list of candidate hostnames, with an optional transformation to apply (ex. replacing dots with dashes). The Agent tries to DNS resolve each combination to ensure its validity before using it as the actual callback it will supply to the Cryostat server. However, when the application that the Agent is attached to is running in Kubernetes and has liveness/readiness probes, then Kubernetes does not set up DNS for the container/Pod until the probes pass - so the Agent's startup setup and attempt to DNS resolve itself will fail early, and never try again. This results in the Agent not being able to register itself with Cryostat when it is attached to applications configured with liveness/readiness probes, which should be a common and expected case.

andrewazores:
Fixes #634

Refactors the callback URI resolution from a one-time startup operation (which is always expected to pass, or else startup bails) to a repeatable operation that is done at the beginning of the registration loop. If the DNS resolution fails then the registration loop fails and will repeat on schedule. This accommodates for cases where the Agent (or its host JVM, or the container they're in) may not be able to resolve its own DNS address early in the startup sequence, for example when there are Kubernetes liveness/readiness probes that must first pass before the Pod can be DNS resolved.


src/main/java/io/cryostat/agent/ConfigModule.java
```diff
@@ -1084,83 +1079,97 @@ private static Optional<URI> buildCallbackFromComponents(Config config) {
...
-                                .build();
-                log.debug("Using {} for callback URL", result.toASCIIString());
-                return Optional.of(result);
-            } catch (URISyntaxException e) {
-                throw new IllegalArgumentException(e);
+        public static CallbackCandidate from(URI uri) {
+            String host, domain;
+            String hostname = uri.getHost();
+            if (hostname.contains(".")) {
+
```
ebaron, line 1102: Unintentional blank line?

ebaron (approved):
Fixes the bug for me, thanks! Just one whitespace issue below.

ebaron (approved):

PR #649, opened 2025-05-27
fix(insights): inherit log level from Cryostat agent

ebaron:
This PR retrieves the `defaultLogLevel` property for the Cryostat agent and applies it to the Insights agent's shaded logger. This can still be overriden by the `RHT_INSIGHTS_JAVA_DEBUG` config parameter to force debug logging for Insights only.

Fixes: https://github.com/cryostatio/cryostat-operator/issues/1090

### DIFF
--- a/src/main/java/io/cryostat/agent/insights/InsightsAgentHelper.java
+++ b/src/main/java/io/cryostat/agent/insights/InsightsAgentHelper.java
@@ -25,6 +25,7 @@ import java.util.function.Supplier;
 
 import io.cryostat.agent.ConfigModule;
 import io.cryostat.agent.model.PluginInfo;
+import io.cryostat.agent.shaded.ShadeLogger;
 
 import com.redhat.insights.agent.AgentBasicReport;
 import com.redhat.insights.agent.AgentConfiguration;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cryostat/agent/insights/InsightsAgentHelper.java
```diff
@@ -67,15 +68,38 @@ public boolean isInsightsEnabled(PluginInfo pluginInfo) {
...
+        // Check if debug logging should be enabled
+        final String defaultLogLevel;
+        boolean debug =
+                config.getOptionalValue(RHT_INSIGHTS_JAVA_DEBUG, boolean.class).orElse(false);
+        if (debug) {
+            out.put("debug", "true");
+            defaultLogLevel = "debug";
+        } else {
+            // Otherwise, apply any log level set for the Cryostat agent
+            defaultLogLevel = System.getProperty(SHADE_PREFIX + SimpleLogger.DEFAULT_LOG_LEVEL_KEY);
```
andrewazores, line 82: There is a class `io.cryostat.agent.shaded.ShadeLogger`, maybe rather than this `SHADE_PREFIX` constant this could use `ShadeLogger.class.getPackageName()`?

ebaron: Nice. I wasn't a fan of hard-coding that.

PR #651, opened 2025-05-28
fix(log): refactor logger to initialize after setting properties

andrewazores:
See #649
See https://github.com/cryostatio/cryostat-operator/pull/1120


ebaron (approved):
Tested in OpenShift with Quarkus and Wildfly test applications, with log levels `off` and `debug`. Works great!

PR #668, opened 2025-07-02
build(maven): update Maven Central publishing URLs

ebaron:
Following the sunset of OSSRH, Sonatype recommends using their [Staging API](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/) to migrate to the Central Publisher Portal to maintain compatibility with existing tools.

Fixes: #669

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -603,8 +603,8 @@
                   <nexus2>
                     <maven-central>
                       <active>ALWAYS</active>
-                      <url>https://s01.oss.sonatype.org/service/local</url>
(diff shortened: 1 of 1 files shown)


andrewazores:
Looks like the new repository doesn't contain any of the old one's contents, so we'd have to move over cryostat-core/libcryostat first?

ebaron:
> Looks like the new repository doesn't contain any of the old one's contents, so we'd have to move over cryostat-core/libcryostat first?

Looks like it. I see releases, but no snapshots: https://central.sonatype.com/artifact/io.cryostat/libcryostat/overview

ebaron:
Looks like the builds pass now

PR #671, opened 2025-07-04
feat(diagnostics): Add endpoint for triggering heap dumps

Josh-Matsuoka:
Fixes https://github.com/cryostatio/cryostat/issues/134

Adds agent support for interacting with the heap dumps system. This allows you to trigger and retrieve heap dumps from a target with the agent enabled. This is done through the mbean-invoke endpoint and the HotspotDiagnosticMBean to create the heap dump, then a similar mechanism to the recording harvester that pushes the heap dump to the server through the heapdump/upload endpoint.

To test:

Setup
- Check out this PR, as well as https://github.com/cryostatio/cryostat-web/pull/1730 and https://github.com/cryostatio/cryostat/pull/984
- Build all 3, build the quarkus-agent sample application with the PR version of the cryostat agent.
- Run smoketest bash smoketest.bash -O -t quarkus-cryostat-agent
- Open the Web UI and test the functionality
- Check the smoketest logs to make sure the agent receives the requests and doesn't throw any errors.

Diagnostics Card

 - Check that the Heap Dumps button appears on the diagnostics card.
 - Verify that clicking the Invoke Heap Dumps button triggers 2 notifications, one for a successful invocation (Heap Dump Success), and one for a successful upload to the server (Heap Dump Uploaded)
 - Verify that the redirect button becomes enabled and takes you to the Heap Dumps table

Capture Diagnostics Page

 - Check that the Heap Dumps button appears on the capture diagnostics page.
 - Verify that clicking the Invoke Heap Dumps button triggers 2 notifications, one for a successful invocation (Heap Dump Success), and one for a successful upload to the server (Heap Dump Uploaded)
 - Verify that the redirect button becomes enabled and takes you to the Heap Dumps table

Heap Dumps Table

 - Check that the earlier triggered heap dumps appear in the table with a name, last modified time, and file size that makes sense
 - Verify that each table entry has a dropdown for downloading and deleting the heap dump.
 - Verify that heap dumps can be downloaded and deleted.
 - Verify that deleting a heap dump produces a notification (Heap Dump Deleted) and refreshes the table


src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -72,6 +82,14 @@ public void handle(HttpExchange exchange) throws IOException {
                                         req.parameters,
                                         req.signature);
 
+                        // TODO: Verify if dumpHeap is blocking, if so we should split the
+                        // invocation
+                        // into a separate thread and listen for when it finishes
+                        if (req.getOperation().equals("dumpHeap")) {
+                            String fileName = req.getParameters()[0].toString();
```
andrewazores, line 89: I think it would be best if the Agent automatically determines the filename that it will dump to, rather than having the server tell it what that path should be. Doing it this way seems like an unnecessary security hole where a request can be sent to the Agent to make it write the heapdump somewhere it should not be on the filesystem (ex. world readable, or a network share, etc.), or else make it write the heapdump overtop of some other important data.

I would suggest that there should be an Agent config property to set a default location for these heap dumps to be written to, and a default file name base/prefix. Appending a UUID to the base name (as is currently done on the server side) seems like a good enough way to mitigate file name conflicts.

andrewazores (approved):

PR #727, opened 2025-10-01
feat(sample): fleet sampling ratio

### LINKED ISSUE #709: [Request] "Fleet Sampling Ratio" config property
When deploying a "fleet" of identical containers (ex. a Deployment with many replica Pods, in k8s terms) it would be useful if the Cryostat Agent had some configuration property to control the odds of the Agent actually starting up and registering with the Cryostat server, starting its Harvester, enabling Smart Triggers, etc. so that a sample subset of the whole fleet population can be monitored. This would be intended for the monitoring/general observability use case more so than profiling or ad-hoc troubleshooting.

The general concept I have is that there would be a config property P with a default value of `1.0`. When the Agent starts, it generates a random number X between 0.0 and 1.0. If `X <= P` then the Agent continues with regular startup as it does now, else it aborts its operations. This way, if the end user will deploy a fleet of 100 containers and only wants to sample approximately 5 of them at any given time, they can set P to 0.05.

Since this is a random value generated on the Agent side, it will not always be a perfect distribution. If the user has some hard constraint about "100 containers, exactly 5 randomly selected for sampling" then there will need to be some external process which orchestrates this. This might be something we can implement using the Cryostat Operator and build on top of its Agent Autoconfig feature. For users not using the Operator, this sampling ratio property might still be of value.

andrewazores:
Fixes #709

### DIFF
--- a/src/main/java/io/cryostat/agent/Agent.java
+++ b/src/main/java/io/cryostat/agent/Agent.java
@@ -239,6 +239,21 @@ public class Agent implements Callable<Integer>, Consumer<AgentArgs> {
         AgentExitHandler agentExitHandler = null;
         try {
             final Client client = DaggerAgent_Client.builder().build();
+
+            boolean sample = client.fleetSampleValue() < client.fleetSamplingRatio();
+            log.trace(
+                    "fleetSampleValue: {} , fleetSampleRatio: {}",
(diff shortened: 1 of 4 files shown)


Josh-Matsuoka (approved):
Tested in a local crc cluster, works nicely

PR #822, opened 2026-02-20
feat(triggers): Call sync endpoint when triggers conclude

### LINKED ISSUE #1329: [Task] SmartTriggers sync endpoint
> Maybe we should add some API endpoint on Cryostat to request it to do a sync of the target. We can surface a UI button to allow the user to do that manually, and when the Agent starts a recording via Smart Trigger it can also call back to Cryostat and ask Cryostat to do a sync of the Agent's own target so that it observes the new recording. What do you think?

This sounds good to me, do you think this is better for a followup PR or in this one?

> Also missing: when the Agent does activate a Smart Trigger and start a recording, and the trigger is removed, there is no notification there either. So the Trigger table stays in its same state displaying a trigger that has already been tripped and removed. This also means deleting a trigger can fail if I'm still viewing that table and the trigger has already been activated by the time I want to delete it, since it's no longer actually present and the view state is stale.

My idea here would be to set up a callback endpoint on the server side /api/beta/smart-triggers/{id}/notify that can be called when a smart trigger completes, sending a removal notification and then doing the recording sync (By calling recordingHelper.syncRecordings if I understand correctly?).

_Originally posted by @Josh-Matsuoka in https://github.com/cryostatio/cryostat/issues/1250#issuecomment-3899276873_
            
Rather than directly calling `recordingHelper.syncRecordings`, I think the new endpoint should schedule a new `TargetUpdateJob` for immediate execution.

There's also some refactoring work to be done so that we can avoid marking these recordings as `external`:

- https://github.com/cryostatio/cryostat/blob/dcd1242722588a620887746ac2c5772eaeca121d/src/main/java/io/cryostat/recordings/RecordingHelper.java#L346
- https://github.com/cryostatio/cryostat/blob/dcd1242722588a620887746ac2c5772eaeca121d/src/main/java/io/cryostat/recordings/ActiveRecording.java#L196

Josh-Matsuoka:
Fixes: https://github.com/cryostatio/cryostat/issues/1329
Related to: https://github.com/cryostatio/cryostat/pull/1357
Depends on: https://github.com/cryostatio/cryostat/pull/1357

Adds support for the new /api/beta/targets/jvmId/smart_triggers/sync/id endpoint in the server. Calls this endpoint to sync the recording and smart trigger state when smart triggers conclude and start recordings.

### DIFF
--- a/src/main/java/io/cryostat/agent/CryostatClient.java
+++ b/src/main/java/io/cryostat/agent/CryostatClient.java
@@ -358,6 +358,18 @@ public class CryostatClient {
         }
     }
 
+    public CompletableFuture<Void> syncSmartTrigger(String id) {
+        HttpPost req =
+                new HttpPost(
+                        baseUri.resolve(
+                                "/api/beta/targets/" + jvmId + "/smart_triggers/sync/" + id));
+        log.trace("{}", req);
+        return supply(req, (res) -> logResponse(req, res))
+                .thenApply(res -> assertOkStatus(req, res))
+                .whenComplete((v, t) -> req.reset())
+                .thenApply(res -> null);
+    }
+
     public CompletableFuture<Void> pushHeapDump(Path heapDump, String requestId)
             throws IOException, IllegalArgumentException {
         Instant start = Instant.now();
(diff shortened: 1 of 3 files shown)


andrewazores:
Looks good, needs rebase

andrewazores (approved):

PR #861, opened 2026-05-01
fix(registration): reuse stored credentials if already in keyring

andrewazores:
Based on #860
Depends on #860

1. Agent checks with Cryostat's credential keyring if the ID of the credential the Agent previously submitted (if any) is still present. If so, the Agent assumes this credential is still valid and does not submit a new one.
2. If the Agent is triggered to enter its registration flow rapidly and a credential generation request is already in-flight, then subsequent requests will reuse a coalesced response from the original in-flight request rather than possibly firing a second request while the first is still in-flight. This also prevents the Agent from rapidly regenerating and invalidating its own webserver credentials.
3. Fixes a byte buffer mutation mishandling where the temporary in-memory webserver password buffer might be zeroed out while the request submitting it to Cryostat's credential keyring is still underway.

### DIFF
--- a/src/main/java/io/cryostat/agent/CredentialCleanupJob.java
+++ b/src/main/java/io/cryostat/agent/CredentialCleanupJob.java
@@ -115,6 +115,21 @@ public class CredentialCleanupJob {
                             .handle(
                                     (v, t) -> {
                                         if (t != null) {
+                                            Throwable cause = t;
+                                            while (cause.getCause() != null) {
+                                                cause = cause.getCause();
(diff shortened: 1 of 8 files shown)


jtolentino1 (dismissed):
LGTM. See: https://github.com/cryostatio/cryostat-agent/pull/860#pullrequestreview-4223212680

andrewazores:
@jtolentino1 needs re-approval after rebase

jtolentino1 (approved):

PR #909, opened 2026-06-09
feat(registration): use simplified registration API

jtolentino1:
Related to cryostatio/cryostat#1494
Depends on cryostatio/cryostat#1586

Uses the combined registration endpoint from cryostatio/cryostat#1586: the Agent registers as a Discovery Plugin, stores its webserver credential, and publishes itself as a target in a single `POST /api/v4.3/discovery/agents` request. This removes the Agent-side credential tracking and cleanup (`CredentialTracker`, `CredentialCleanupJob`, and all `/api/v4/credentials` / `credential_exists` interactions), so after registration the Agent's internal state is just its webserver credentials and its Discovery Plugin ID/token. The plugin JWT is used for the periodic registration check and for deregistration on clean shutdown, which now relies on the server to cascade-delete the associated Stored Credential.


src/main/java/io/cryostat/agent/CryostatClient.java
```diff
@@ -150,218 +144,46 @@ public CompletableFuture<Boolean> checkRegistration(PluginInfo pluginInfo) {
...
-                                    credentialTracker.markForDeletion(credentialId);
                                     throw new RegistrationException(e);
                                 }
                             })
-                    .whenComplete((v, t) -> req.reset());
+                    .whenComplete((v, t) -> req.reset())
+                    .whenComplete((v, t) -> Arrays.fill(passwordCopy, (byte) 0));
         } catch (JsonProcessingException e) {
-            credentialTracker.markForDeletion(credentialId);
+            Arrays.fill(passwordCopy, (byte) 0);
```
andrewazores: I suppose this would be better in a `finally` so that it isn't specific to the `JsonProcessingException`

src/main/java/io/cryostat/agent/CryostatClient.java
```diff
@@ -616,30 +438,75 @@ private ClassicHttpResponse assertOkStatus(HttpUriRequestBase req, ClassicHttpRe
...
+
+        public Map<String, String> getContext() {
+            return context;
+        }
+    }
+
+    static class AgentCredential {
+        private final String matchExpression;
+        private final String username;
+        private final String password;
```
andrewazores: Can this password field be `byte[]`? This is used elsewhere when dealing with the password so that the buffer can be cleared explicitly and the contents aren't just left sitting around on heap until the GC decides to run. If that byte buffer gets copied into an immutable structure like a String then we're back to being at the mercy of the GC.

src/main/java/io/cryostat/agent/CryostatClient.java
```diff
@@ -150,218 +144,46 @@ public CompletableFuture<Boolean> checkRegistration(PluginInfo pluginInfo) {
     }
 
     public CompletableFuture<PluginInfo> register(
-            int credentialId, PluginInfo pluginInfo, URI callback) {
+            URI callback, CredentialsSnapshot credentials, Collection<DiscoveryNode> subtree) {
+        byte[] passwordCopy = Arrays.copyOf(credentials.pass(), credentials.pass().length);
```
andrewazores: `CredentialsSnapshot#pass` already creates a copy, which in this case is done twice and each of those copies is not managed and cleared and relies on the GC to discard it, too. It looks like `byte[] passwordCopy = credentials.pass()` would be fine.

src/main/java/io/cryostat/agent/CryostatClient.java
andrewazores: The reason for this is just so that the plaintext password content is only on heap for a very short time (only as long as necessary for operation), so the risk of it getting captured by a heap dump and potentially leaked is minimized. Once the plaintext password has been sent off to Cryostat, then Cryostat retains a copy in its encrypted database keyring, and the Agent only keeps a hash of the original password on its own heap.

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -480,5 +443,30 @@ private byte[] hash(String pass) {
...
+            if (passHash.length == 0) {
+                throw new IllegalStateException("credentials have not been generated");
+            }
+            return new CredentialsSnapshot(user, Arrays.copyOf(pass, pass.length));
+        }
+    }
+
+    static class CredentialsSnapshot {
+        private final String user;
+        private final byte[] pass;
```
andrewazores, line 469: Trying back to the other thread about password contents on heap, this field needs a way to be explicitly cleared as well, and whatever callsites are using these snapshots needs to ensure they clear when they're done.

Maybe making this `AutoCloseable` could be a good marker to remind callers that this should be done, too.

src/main/java/io/cryostat/agent/Registration.java
```diff
@@ -347,22 +315,25 @@ void tryRegister() {
...
-                                        return cryostat.register(
-                                                credentialId, pluginInfo, credentialedCallback);
-                                    } catch (URISyntaxException e) {
+                                        Set<DiscoveryNode> selfNodes = defineSelf();
+                                        log.trace(
+                                                "registering and publishing self as {}",
+                                                selfNodes.stream()
+                                                        .map(n -> n.getTarget().getConnectUrl())
+                                                        .collect(Collectors.toList()));
+                                        WebServer.CredentialsSnapshot credentials =
```
andrewazores:
```java
try (var credentials = webServer.getCredentialsSnapshot()) {
 return cryostat.register(...);
}
```

src/main/java/io/cryostat/agent/WebServer.java
andrewazores, line 390: Maybe this should also be `AutoCloseable` to also mark for callers that it should be closed/cleared.

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -480,5 +444,37 @@ private byte[] hash(String pass) {
         private byte[] hash(byte[] bytes) {
             return digest.digest(bytes);
         }
+
+        synchronized CredentialsSnapshot snapshot() {
+            if (passHash.length == 0) {
+                throw new IllegalStateException("credentials have not been generated");
+            }
+            return new CredentialsSnapshot(user, Arrays.copyOf(pass, pass.length));
```
andrewazores: Once this snapshot is created, I think the internal plaintext state can be cleared, right? Since the new snapshot is going to hold that state and produce copies, so this original Credentials structure just needs to track the hash from this point on.

andrewazores:
```java
var snapshot = new CredentialsSnapshot(user, copyOfPass);
close();
return snapshot;
```

andrewazores: Or if Credential is made closeable, I wonder if this syntax is legal:

```java
try (this) {
  return new CredentialsSnapshot(user, copyOfPass)
}
```

andrewazores: Actually, nevermind this - on further thought, the Credential shouldn't be overall closed because it's not necessarily a one-time resource. If the credentials are regenerated during runtime then it doesn't make sense for the Credential to have been "closed" previously.

Clearing the internal plaintext password immediately after creating the copyOfPass and snapshot still makes sense I think, just not the AutoCloseable part.

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -480,5 +444,37 @@ private byte[] hash(String pass) {
...
+        }
+    }
+
+    @SuppressFBWarnings("EI_EXPOSE_REP2")
+    static class CredentialsSnapshot implements AutoCloseable {
+        private final String user;
+        private final byte[] pass;
+
+        // takes ownership of the pass buffer, which is zeroed on close()
```
andrewazores: This says it takes ownership of the buffer, but the one place that this is instantiated passes in a copy anyway. Maybe this could be inverted, the copying can take place in this constructor? Then it's more clear that the buffer ownership doesn't actually transfer to the snapshot (and multiple snapshots could be made, if needed in the future) since each snapshot creates its own copy of the buffer and should be closed out when no longer needed, clearing out that individual copy.

src/main/java/io/cryostat/agent/WebServer.java
andrewazores: I think this is no longer used now that there's the CredentialsSnapshot.

src/main/java/io/cryostat/agent/CryostatClient.java
```diff
@@ -59,25 +62,23 @@
...
 import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.io.entity.ByteArrayEntity;
 import org.apache.hc.core5.http.io.entity.StringEntity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class CryostatClient {
 
     private static final String DISCOVERY_API_PATH = "/api/v4/discovery";
+    private static final String AGENT_REGISTRATION_API_PATH = "/api/v4.3/discovery/agents";
```
andrewazores, line 86: #907 - it's probably worthwhile to at least add a comment about this set of constants pointing authors to the Operator and the need to update the nginx conf.

src/main/java/io/cryostat/agent/WebServer.java
```diff
@@ -449,19 +447,22 @@ synchronized CredentialsSnapshot snapshot() {
             if (passHash.length == 0) {
                 throw new IllegalStateException("credentials have not been generated");
             }
-            return new CredentialsSnapshot(user, Arrays.copyOf(pass, pass.length));
+            if (plaintextCleared) {
+                throw new IllegalStateException("plaintext credentials have already been cleared");
```
andrewazores, line 459: I think the `plaintextCleared` field can be avoided if this just checks if the `pass` is equal to an all-null-byte array, like `if (pass.equals(new byte[pass.length]))`, or without allocation by checking each `pass[i] == 0`. This just avoids adding a bit of derived state which can already be determined cheaply by examining the other existing state.

andrewazores (approved):

PR #914, opened 2026-06-20
feat(graphQL): Add jvmId to AsyncProfiler response

Josh-Matsuoka:
Related to: [ #1254](https://github.com/cryostatio/cryostat/issues/1254)

Required for:

Adds the JvmID to the asyncProfile object for use in the graphQL query on the server side.

### DIFF
--- a/src/main/java/io/cryostat/agent/remote/AsyncProfilerContext.java
+++ b/src/main/java/io/cryostat/agent/remote/AsyncProfilerContext.java
@@ -72,17 +72,20 @@ class AsyncProfilerContext extends MutatingRemoteContext {
     private final ScheduledExecutorService scheduler;
     private final Logger log = LoggerFactory.getLogger(getClass());
     private volatile StartProfileRequest currentProfile = null;
+    private final String jvmId;
 
     @Inject
     AsyncProfilerContext(
             SmallRyeConfig config,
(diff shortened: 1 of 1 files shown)


src/main/java/io/cryostat/agent/remote/AsyncProfilerContext.java
```diff
@@ -72,17 +72,20 @@ class AsyncProfilerContext extends MutatingRemoteContext {
...
     private final Logger log = LoggerFactory.getLogger(getClass());
     private volatile StartProfileRequest currentProfile = null;
+    private final String jvmId;
 
     @Inject
     AsyncProfilerContext(
             SmallRyeConfig config,
             ObjectMapper mapper,
             ScheduledExecutorService scheduler,
+            @Named("JVM_ID") String jvmId,
```
andrewazores, line 82: This should be `@Named(MainModule.JVM_ID)`, ie using the defined constant (which will need to be made visible).

But what's the reason that the AsyncProfile struct needs to contain the jvmId? I notice that the equivalent struct in RecordingsContext doesn't do this. It seems like from the GraphQL side, Cryostat must already know which target instance it's referring to (and therefore its JVM ID) before it's going to send an Agent API request to do some async-profiler action.

PR #953, opened 2026-07-13
feat(triggers): Smart Triggers JSON prototype

### LINKED ISSUE #817: [Request] Remove Smart Triggers command line argument syntax and use JSON serialization
The `[expr ; TargetDuration > n]~template` Smart Trigger syntax should be removed along with the facility for passing these "raw" trigger definitions to the Agent via command line argument.

Instead, Smart Triggers should be created by loading from file or by remote management (both of which are already supported), and both of these should use JSON serialization of trigger definitions instead of the "square bracket, semicolon, tilde" syntax.

Josh-Matsuoka:
Fixes: https://github.com/cryostatio/cryostat-agent/issues/817

Depends on: https://github.com/cryostatio/cryostat-core/pull/747 https://github.com/cryostatio/cryostat/pull/1736 https://github.com/cryostatio/cryostat-web/pull/2411

Adds a prototype implementation of smart triggers as a json representation, for example:

{ condition : ProcessCpuLoad>0.1, duration: 30s, template: profiling }

This is easily extensible later to support e.g. an operation parameter specifying what kind of action to take when the trigger conditions are met. This fully abstracts away the CEL implementation from the user, all they need to supply is the conditions, durations, and templates. We can construct any needed expressions from there.

UUIDs are not included in the request representation, as they're generated internally rather than by the user. This means to make use of the delete endpoint the user will still need to call GET first to retrieve the representation with the IDs listed.

Since this is a breaking change I've contained it for now to its' own endpoint /smart-triggers-beta/ and config option (CRYOSTAT_AGENT_SMART_TRIGGER_BETA_FORMAT).

Opening as draft while tests are added


src/main/java/io/cryostat/agent/remote/SmartTriggersBetaContext.java
```diff
@@ -0,0 +1,157 @@
...
+                    // Delete is unchanged, it doesn't need the full JSON Construction, just the ID
+                    try (InputStream body = exchange.getRequestBody()) {
+                        // UUID is passed as a path param
+                        Matcher m = PATH_ID_PATTERN.matcher(exchange.getRequestURI().getPath());
+                        if (!m.matches()) {
+                            exchange.sendResponseHeaders(
+                                    HttpStatus.SC_BAD_REQUEST, BODY_LENGTH_NONE);
+                        }
+                        String uuid = m.group(1);
+                        log.trace("Extracted uuid: " + uuid);
```
andrewazores: nitpick: should use the `{}` templating, ie `("Extracted uuid: {}", uuid)`

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -42,15 +44,21 @@ public class TriggerParser {
             "\\[(.*(&&)*|(\\|\\|)*)\\]~([\\w\\-]+)(?:\\.jfc)?";
     private static final Pattern EXPRESSION_PATTERN = Pattern.compile(EXPRESSION_PATTERN_STRING);
     private final FlightRecorderHelper flightRecorderHelper;
+    private ObjectMapper mapper;
```
andrewazores: This looks like it could be `final`

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -134,4 +156,57 @@ public boolean isValid(String definition) {
...
     }
+
+    public List<SmartTrigger> parseFromJson(String req) {
+        try {
+            SmartTriggerReq[] reqs = mapper.readValue(req, SmartTriggerReq[].class);
+            var returnVal = new ArrayList<SmartTrigger>();
+            for (SmartTriggerReq r : reqs) {
+                // Null/non-provided fields will already be caught by the
+                // ObjectMapper, check their values are valid
+                System.out.println(
```
andrewazores: Remove?

src/main/java/io/cryostat/agent/remote/SmartTriggersContext.java
```diff
@@ -60,26 +61,33 @@ public void handle(HttpExchange exchange) throws IOException {
...
                         exchange.sendResponseHeaders(HttpStatus.SC_OK, BODY_LENGTH_UNKNOWN);
                         try (OutputStream responseStream = exchange.getResponseBody()) {
-                            mapper.writeValue(responseStream, respUUID);
+                            mapper.writeValue(responseStream, triggerIds);
                         }
                     } catch (Exception e) {
                         log.warn("Smart trigger serialization failure", e);
                         exchange.sendResponseHeaders(HttpStatus.SC_BAD_GATEWAY, BODY_LENGTH_NONE);
                     }
                     break;
```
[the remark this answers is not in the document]
Josh-Matsuoka: @andrewazores since we're making the breaking change, the first comment here isn't relevant, right?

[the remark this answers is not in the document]
andrewazores: Agreed, it seems @coderabbitai is confused by the sequence of events in the commit history. There is no beta API here (anymore) - this is an intentional breaking change of the Smart Triggers serialization between Cryostat and the Agent.

Josh-Matsuoka:
dependent issues check failing due to server issues: 

>   Error: HttpError: No server is currently available to service your request. Sorry about that. Please try resubmitting your request and contact us if the problem persists.

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    // The CEL internal representation doesn't need to be exposed
+    // to users, we can construct the expression to evaulate
+    // from a simple set of properties.
+    public String constructExprFromParams() {
+        return this.condition + constructDurationExprFromRequest();
+    }
+
+    public String constructDefinitionFromParams() {
```
andrewazores: What is this method used for? Is it possible to refactor deeper so that this old `[foo]~template` syntax is eliminated entirely? The `SmartTriggerReq` should be usable not only as a (de)serialization JSON format for communication between the Agent and Cryostat, but it should also be a sufficient internal representation of what the trigger is/does for the trigger evaluator and other internal mechanisms.

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+public class SmartTriggerReq {
+
+    // TODO: For now the only supported operation is starting a recording,
+    // if/when work proceeds on supporting e.g. thread/heap dumps this can
+    // be extended to support an operation type.
+    // { condition, durationExpr, template }
+    private String condition;
+    private String durationExpr;
```
andrewazores: Similar to what I said over here: https://github.com/cryostatio/cryostat-web/pull/2411#issuecomment-5318702739

Rather than having a `String durationExpr` field here, could this now be represented as a numeric type? For example, this could be a `long` which is interpreted as milliseconds. That's really the key piece of information that needs to be conveyed between the user, server, and agent. The specifics of the CEL expression for `TargetDuration>duration("$x")` is the kind of internal implementation detail we're trying to stop exposing, and I think all of those pieces are why this field is currently a `String`.

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
Josh-Matsuoka: This can be moved to the TriggerParser and be done there, the idea was to keep the existing parsing infrastructure in place, the JSON format already contains all the relevant information (Condition, duration, template), this just builds that expression to be passed to the parser and turned into the CEL format used by the evaluator.

andrewazores: Right, but do we even need a parser anymore?

andrewazores: For example, there's a line like this in the parser:

```java
                    returnVal.addAll(parse(constructDefinitionFromParams(r)));
```

Where `returnVal` is a `List<SmartTrigger>` and `r` is a `SmartTriggerReq`. Can't we transform a `SmartTriggerReq` into a `SmartTrigger` without using the `[foo]~template` String syntax as an intermediary?

Josh-Matsuoka: There's a bit of cleanup that can be done to simplify the parser, but SmartTrigger still expects a certain format. We can pass foo and template, but we'd still need to construct the duration expression (targetDuration<duration(x)) for foo.

andrewazores: Sure, the `TargetDuration<duration(x)` is still needed for the CEL duration comparison as an internal detail. That part is fine now that it is an internal detail and the user only knows about the `x` part.

If the cryostat-core/libcryostat SmartTrigger structure also needs to be adjusted to clean up its interface then that's also worth pursuing.

Josh-Matsuoka: I've simplified the parser to construct the SmartTriggers directly instead of parsing a constructed String, the libcryostat SmartTrigger structure is fine as-is I think.

src/main/java/io/cryostat/agent/remote/SmartTriggersContext.java
```diff
@@ -60,19 +60,17 @@ public void handle(HttpExchange exchange) throws IOException {
...
                 case "POST":
                     try (InputStream body = exchange.getRequestBody()) {
-                        SmartTriggerRequest req = mapper.readValue(body, SmartTriggerRequest.class);
-                        List<String> respUUID = evaluator.append(req.definitions);
+                        SmartTriggerReq[] reqs = mapper.readValue(body, SmartTriggerReq[].class);
                         exchange.sendResponseHeaders(HttpStatus.SC_OK, BODY_LENGTH_UNKNOWN);
                         try (OutputStream responseStream = exchange.getResponseBody()) {
-                            mapper.writeValue(responseStream, respUUID);
+                            mapper.writeValue(responseStream, evaluator.append(reqs));
```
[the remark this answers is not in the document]
andrewazores: Disagree, returning UUIDs from POST is fine here. The client should not be required to perform a GET.

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
```diff
@@ -0,0 +1,64 @@
...
+        this.condition = "";
+        this.recordingTemplate = "";
+    }
+
+    public long getDuration() {
+        return duration;
+    }
+
+    public void setDuration(long duration) {
+        this.duration = duration;
```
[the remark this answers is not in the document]
andrewazores: If the field uses boxed Long instead of primitive long so that it's nullable, does that convey the "optional duration" intent?

src/main/java/io/cryostat/agent/remote/SmartTriggersContext.java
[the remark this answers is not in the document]
Josh-Matsuoka: This isn't a problem, GET can be used to retrieve the IDs of existing triggers for removal, POST is fine returning the UUIDs of appended triggers

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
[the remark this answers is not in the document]
andrewazores: I think the intent is that the duration is optional in the sense that a trigger can be activated immediately as soon as its condition is met. This could be encoded as a null duration field, but I think a non-positive duration field could also capture this intent as well, so long as this is properly documented.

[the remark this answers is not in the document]
andrewazores: @Josh-Matsuoka something minor to document here.

Josh-Matsuoka:
On second thought it is better to simplify the targetDuration handling of the SmartTrigger structure. This lets us remove the expression generation entirely, the only relevant one is the duration expression which can be created by the SmartTrigger object instead.

CI checks expected to fail due to libcryostat changes

https://github.com/cryostatio/cryostat-core/pull/747

src/main/java/io/cryostat/agent/triggers/TriggerParser.java
```diff
@@ -92,46 +96,96 @@ private boolean checkDir() {
...
+
+    // Blank Duration indicates the trigger should fire immediately
+    // when the condition is met.
+    private String constructDurationExprFromRequest(SmartTriggerReq req) {
+        return req.getDuration() == 0
+                ? ""
+                : ";TargetDuration>duration(\""
+                        + Duration.ofMillis(req.getDuration()).toSeconds()
+                        + "s\")";
+    }
```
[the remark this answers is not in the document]
Josh-Matsuoka: Resolved by libcryostat changes

src/main/java/io/cryostat/agent/triggers/SmartTriggerReq.java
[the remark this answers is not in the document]
Josh-Matsuoka: omitted duration is fine, it's treated as a trigger that should fire immediately when conditions are met.

andrewazores:
Not sure if the root cause is on the Agent side, or server/frontend, but when I try to create a trigger now via UI I get an HTTP 502 response and this appears in the Agent's logs:

```
2026-08-19 17:31:53:581 +0000 [cryostat-agent-worker-2] WARN io.cryostat.agent.remote.SmartTriggersContext - Smart trigger serialization failure
io.cryostat.agent.shaded.com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `io.cryostat.agent.triggers.SmartTriggerReq` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:72)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1814)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1588)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1535)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeFromArray(BeanDeserializer.java:692)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeOther(BeanDeserializer.java:203)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:179)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:5052)
        at io.cryostat.agent.shaded.com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3991)
        at io.cryostat.agent.remote.SmartTriggersContext.handle(SmartTriggersContext.java:71)
        at io.cryostat.agent.WebServer.lambda$wrap$0(WebServer.java:232)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:95)
        at jdk.httpserver/sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:71)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:98)
        at jdk.httpserver/sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:928)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:95)
        at io.cryostat.agent.WebServer$RequestLoggingFilter.doFilter(WebServer.java:291)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:98)
        at io.cryostat.agent.WebServer$CooldownFilter.doFilter(WebServer.java:323)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:98)
        at jdk.httpserver/sun.net.httpserver.ServerImpl$Exchange.run(ServerImpl.java:902)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:840)
```

Josh-Matsuoka:
I can't reproduce that, are you sure you've updated and built all of the PRs? With everything updated and built together I'm seeing normal behavior from the UI.

Trigger UI

andrewazores:
Hmm, let me try again. I did try to rebuild everything locally but maybe I missed a step, or maybe one of Cryostat or the Agent used an outdated -core.

andrewazores (approved):

PR #973, opened 2026-07-22
feat(unified-log): implement JVM Unified Logging management and retrieval

andrewazores:
Fixes #825
See cryostatio/cryostat#1710


src/main/java/io/cryostat/agent/remote/UnifiedLogging.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+        return parseVmLogListOutput(output);
+    }
+
+    State parseVmLogListOutput(String output) {
+        String lastFileLine = null;
+        String lastActiveNonFileLine = null;
+        for (String line : output.split("\n")) {
+            if (VM_LOG_LIST_FILE_PATTERN.matcher(line).find()) {
+                lastFileLine = line;
```
jtolentino1, line 119: Should this filter for a GC-enabled output rather than taking the last file entry? I tried a JVM with both -`Xlog:gc:file=/tmp/gc.log` and `-Xlog:class+load:file=/tmp/class.log`. `VM.log list` put the class-load output last, so this reported `/tmp/class.log` as the GC log.

src/main/java/io/cryostat/agent/remote/InvokeContext.java
```diff
@@ -138,6 +161,43 @@ public void handle(HttpExchange exchange) throws IOException {
...
+            req.parameters[0] = col.toArray(new String[0]);
+        }
+        String vmLogArgs;
+        if (req.parameters[0] instanceof String[]) {
+            String[] strArr = (String[]) req.parameters[0];
+            vmLogArgs = strArr.length > 0 ? strArr[0] : "";
+        } else {
+            vmLogArgs = String.valueOf(req.parameters[0]);
+        }
+        if (!vmLogArgs.contains("disable=true") && !vmLogArgs.contains("rotate")) {
```
jtolentino1, line 188: I think `VM.log disable=true` clears all unified logging ... can we do `output=<gc path> what=all=off` to only disable the GC output.

src/main/java/io/cryostat/agent/remote/UnifiedLogging.java
```diff
@@ -0,0 +1,270 @@
...
+
+    /** Immutable snapshot of the JVM unified logging configuration for GC output. */
+    static class State {
+        public final boolean enabled;
+
+        @JsonSerialize(using = ToStringSerializer.class)
+        public final Path logFilePath;
+
+        public final String what;
+        public final String decorators;
```
jtolentino1, line 67: Need `hasLog` here

https://github.com/cryostatio/cryostat/pull/1710/changes#diff-848f2935015382d8aa8c1f41b50afb8f639e45a7bedc8a4ec1bc5650775eab7bR905

jtolentino1 (dismissed):

src/main/java/io/cryostat/agent/remote/UnifiedLogging.java
andrewazores: Actually, I just removed `hasLog` on the server's side. That isn't used anymore.

jtolentino1: Ah okay, gotcha

src/main/java/io/cryostat/agent/remote/UnifiedLogging.java
andrewazores: I think this actually points to the feature being misnamed. The original feature request was specific to GC log collection, but on modern (HotSpot) JVMs the GC log configuration is tied in with lots of other kind of logging via the unified logging system. Rather than restricting the scope and usage of the feature to only handle GC logs, maybe I should just rename this feature to "Logging" or "Unified Logging" more generally.

src/main/java/io/cryostat/agent/remote/InvokeContext.java
andrewazores: This isn't invoking a `disable=true`, it's checking if that's what's being invoked so that it can pass through disable (and rotate) commands as-is. Other commands, ones for reconfiguring logging, are intercepted here and modified to ensure that the `output` is set to the file path the Agent needs it to be, and `output_options` is configured according to the Agent's configuration (so neither of those can be set by the remote client).

src/main/java/io/cryostat/agent/remote/UnifiedLogging.java
jtolentino1: That makes sense to me, renaming around unified logging sounds more accurate

src/main/java/io/cryostat/agent/remote/InvokeContext.java
jtolentino1: Oh got it, thanks for clarifying.

jtolentino1 (approved):

